PR #23, opened 2015-03-20
Fix bug in localintersection2by2

statik:
I think there was a problem with localintersection2by2. Added a failing test case to illustrate the problem, and a code change that passes the new test case and the existing test suite.

### DIFF
--- a/setutil.go
+++ b/setutil.go
@@ -219,7 +219,7 @@ mainwhile:
 				if k2 == len(set2) {
 					break mainwhile
 				}
-				if set2[k2] <= set1[k1] {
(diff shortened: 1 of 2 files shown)


lemire:
Thanks.  Normally our unit tests should have caught such a bug. I will have a closer look soon to figure out what happened.

lemire:
Ok. I am very puzzled as to how this code could go unchecked for so long. It appears to be nearly identical to the Java code (a straight port) yet the inequalities have been reversed for some reason (which is obviously wrong).

Of course, the solution is to immediately merge your PR.

PR #25, opened 2015-04-08
Implement Select

bpot:
Roughly follows the same implementation as the Java library.

The method on the containers can't be named `select` because it conflicts with the `select` keyword. I'm happy to change it to something better than `selectInt` but I couldn't come up with anything.

I noticed that the other methods on RoaringBitmap don't return an `error` value. I added one for `Select` since I didn't like the idea of having the method 'silently' fail by returning `-1` if it was given an input it couldn't handle.

I haven't done any benchmarking of `selectBitPosition` but I think it should be reasonably efficient.

### DIFF
--- a/arraycontainer.go
+++ b/arraycontainer.go
@@ -393,6 +393,10 @@ func (ac *arrayContainer) rank(x uint16) int {
 	}
 }
 
+func (ac *arrayContainer) selectInt(x uint16) int {
+	return int(ac.content[x])
+}
+
 func (ac *arrayContainer) clone() container {
 	ptr := arrayContainer{make([]uint16, len(ac.content))}
 	copy(ptr.content, ac.content[:])
--- a/bitmapcontainer.go
+++ b/bitmapcontainer.go
@@ -361,6 +361,18 @@ func (bc *bitmapContainer) rank(x uint16) int {
 	}
 }
 
+func (bc *bitmapContainer) selectInt(x uint16) int {
+	remaining := x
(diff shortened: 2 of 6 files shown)


lemire:
Thanks. I will review it soon.

Users should never use the containers directly, so as long as we, as insiders, can find our way, names are not super important.

lemire:
Your  `selectBitPosition` is definitively faster and better than what I had in Java, so I have ported your algorithm to the Java library after benchmarking it:

https://github.com/lemire/RoaringBitmap/commit/887417159bece7546eeba299a4695ab2e8c43a7a

You are credited.

lemire:
@bpot 

However, as it stands, your `selectBitPosition` is broken. (This suggests that our unit tests are too weak.)

Here is a demo showing the problem: (click `run`)

https://play.golang.org/p/sRs7atK933

lemire:
@bpot  Please have a look at my proposed fixes.

bpot:
@lemire those changes look good to me! I've updated the PR. nice catch.

Slight tangent, for testing I think it may be useful to do some kind of generative property based testing (like haskell's quickcheck) instead of  (or in addition to) enumerating different edge cases. Haven't fully thought through what this might look like but seems like it would be a good fit.

lemire:
@bpot 

The unit tests are a bit of a mess and they are not sufficient. Any improvement would help.

lemire:
@bpot  I am told you should have write access now to the repository.

PR #26, opened 2015-04-09
Make Clone() a deep clone

bpot:
The current method is doing a shallow copying of just the references to the elements in the roaring array.

This was causing panics during iteration because the parent bitmap could end up with an invalid structure (element with no values set).


lemire:
Thanks. I will review this shortly.

lemire:
Looks like a no-brainer so I am merging without further review.

Thanks.

Issue #31, opened 2015-06-11
Reduce memory usage by getting rid of the element struct


lemire:
The element struct means that our 16-bit keys probably consume a lot more than 16 bits. The code should be rewritten without an element struct for reduced memory usage.

lemire:
Implemented by @bpot

PR #32, opened 2015-06-22
Remove element type and use two parallel arrays for key/container

bpot:
As described in #31 

For a [program](https://gist.github.com/bpot/49455d04f24da7e80582) which sets a single bit for each high-level key it reduces space usage by about 15% and object count by 40%.

Benchmarks appear unchanged.

### Master

```
HeapInUse 6072410112
HeapObjects 163851038
```

### Without element

```
HeapInUse 5198454784
HeapObjects 98319536
```


lemire:
@bpot  Excellent!

lemire:
@bpot  I have added a version of your little program as part of our benchmarks... (though I allocate far fewer objects... for speed reasons)

In the future, if one can reduce memory usage, this could serve as a benchmark.

PR #33, opened 2015-06-28
Reduce duplication in roaringarray

bpot:


lemire:
Can't find anything to complain about, so merging.

PR #35, opened 2015-06-30
Copy-on-write optimization for Clone()

bpot:
With this change a cloned bitmap will initially share the containers with its parent bitmap. If there are changes to a container in the parent or clone we duplicate the container before modifying it. This PR implements this as part of the existing `Clone` method but we could create a new method to make this behavior opt-in (although I think in most cases the new behavior will be preferable).

This functionality adds an additional 24 bytes for the dirty slice and an additional byte per container to track the dirty state (if the `Clone` method is used).

## Benchmarks

There is a small 1-2% performance regression for `Add`.

I can't account for the changes to the sparse iterate and intersection benchmarks. Especially `BenchmarkSparseIterateBitset` which doesn't seem to use the roaring package at all. The comparison below uses the best of 5 runs for each benchmark and the results are pretty consistent. We may be seeing something like https://github.com/golang/go/issues/8717

```
benchmark                         old ns/op     new ns/op     delta
BenchmarkSparseIterateBitset      2749933       2925965       +6.40%
BenchmarkSparseIterateRoaring     967890        911762        -5.80%
BenchmarkIntersectionRoaring      2280          2413          +5.83%
BenchmarkSetBitset                28.7          28.2          -1.74%
BenchmarkUnionRoaring             460354        468094        +1.68%
BenchmarkSetRoaring               77.5          78.5          +1.29%
BenchmarkUnionBitset              6062277       6010689       -0.85%
BenchmarkGetTestBitSet            25.8          25.6          -0.78%
BenchmarkIntersectionBitset       11323         11301         -0.19%
BenchmarkIterateBitset            646494        647617        +0.17%
BenchmarkIterateRoaring           1563493       1565886       +0.15%
BenchmarkCountBitset              31716         31710         -0.02%
BenchmarkCountRoaring             74.5          74.5          +0.00%
BenchmarkGetTestRoaring           155           155           +0.00%
```

### DIFF
--- a/roaring.go
+++ b/roaring.go
@@ -166,15 +166,22 @@ func (rb *RoaringBitmap) Equals(o interface{}) bool {
 // Add the integer x to the bitmap
 func (rb *RoaringBitmap) Add(x uint32) {
 	hb := highbits(x)
-	i := rb.highlowcontainer.getIndex(hb)
+	ra := rb.highlowcontainer
+	i := ra.getIndex(hb)
 	if i >= 0 {
-		rb.highlowcontainer.setContainerAtIndex(i, rb.highlowcontainer.getContainerAtIndex(i).add(lowbits(x)))
+		var c container
+		if ra.hasDirty() {
+			c = ra.getWritableContainerAtIndex(i)
+		} else {
+			c = ra.getContainerAtIndex(i)
+		}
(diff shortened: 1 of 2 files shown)


lemire:
@bpot  Sorry for the delay, merged just now.

PR #36, opened 2015-07-10
Fix CheckedRemove

boutros:
Bitmap cardinality should be _less_ than old cardinality if an integer was removed


lemire:
Duh! Thanks!

PR #39, opened 2015-09-23
methods to serialize and deserialize from base 64 encoded string

tvmaly:
Please consider the pull request for the methods  ToBase64 and FromBase64.  I have provided a test case.  I currently use the feature with a web service that stores the roaring bitmaps in a redis key value store.

### DIFF
--- a/roaring.go
+++ b/roaring.go
@@ -7,6 +7,7 @@ package roaring
 
 import (
 	"bytes"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"strconv"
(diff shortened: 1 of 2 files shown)


lemire:
@tvmaly  Code looks good. Merging.

Do you have a simple code sample (for documentation purposes) of the use of roaring in redis?

lemire:
@tvmaly  I will be adding your name to the list of contributors (in the README).

tvmaly:
Professor, for a redis sample,  assuming usage of the library "github.com/garyburd/redigo/redis"
and a connection to redis in variable conn, and the key somekey where the string is stored it would be as follows:
	roaring_string, err := redis.String(conn.Do("GET", somekey))
        rb := roaring.NewRoaringBitmap()

			if roaring_string != "" {
				_, err = rb.FromBase64(roaring_string)
				if err != nil {
					log.Printf("Error decoding key %s base64 to roaring: %s\n", somekey, err)
					return err
				}
			}

lemire:
@tvmaly 

Excellent... do you have a code sample to store an existing roaring object in redis?

lemire:
@tvmaly  The goal would be to add to the documentation so that other users who are interested in using redis know right away how to call your contributed methods. (I find that people really like concrete code examples.)

PR #41, opened 2015-10-07
Use unsafe/reflect in serialization

willglynn:
Serialized representations are just little endian byte slices, which can be obtained by referencing the data slice directly on certain platforms. 

On my machine, this results in 64%, 88%, and 90% reductions in runtime for `BenchmarkSerialization{Sparse,Mid,Dense}`:

    $ git checkout master
    Switched to branch 'master'
    Your branch is up-to-date with 'origin/master'.
    $ go test -bench BenchmarkSerialization -run -
    PASS
    BenchmarkSerializationSparse-8	    5000	    275510 ns/op
    BenchmarkSerializationMid-8   	   10000	    128213 ns/op
    BenchmarkSerializationDense-8 	  100000	     20826 ns/op
    ok  	github.com/tgruben/roaring	5.112s
    $ git checkout unsafe_serialization
    Switched to branch 'unsafe_serialization'
    $ go test -bench BenchmarkSerialization -run -
    PASS
    BenchmarkSerializationSparse-8	   20000	     97988 ns/op
    BenchmarkSerializationMid-8   	  100000	     14870 ns/op
    BenchmarkSerializationDense-8 	 1000000	      2043 ns/op
    ok  	github.com/tgruben/roaring	6.800s

### DIFF
--- a/arraycontainer.go
+++ b/arraycontainer.go
@@ -1,21 +1,9 @@
 package roaring
 
-import (
-	"encoding/binary"
-	"io"
-	"unsafe"
-)
+import "unsafe"
 
 type arrayContainer struct {
 	content []uint16
-}
-
-func (b *arrayContainer) readFrom(stream io.Reader) (int, error) {
-	err := binary.Read(stream, binary.LittleEndian, b.content)
-	if err != nil {
-		return 0, err
-	}
-	return 2 * len(b.content), nil
 }
 
 func (ac *arrayContainer) fillLeastSignificant16bits(x []uint32, i int, mask uint32) {
--- a/arraycontainer.go
+++ b/arraycontainer.go
@@ -10,17 +10,6 @@ type arrayContainer struct {
 	content []uint16
 }
 
-// writes the content (omitting the cardinality)
-func (b *arrayContainer) writeTo(stream io.Writer) (int, error) {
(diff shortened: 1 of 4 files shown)


lemire:
@willglynn  

Excellent.

Before merging this... one wonders... is there any chance that it would be easy to do the same thing for serialization, so that we have both accelerated serialization and deserialization?

Anyhow, well worth your name on the contributors' list.

willglynn:
Hmm… I don't see why not. `readFrom()` gets an `io.Reader`, so we can copy straight into a caller-owned buffer, which we can reinterpret straight into a `uint16`/`uint64` slice on little-endian platforms. Expect a second commit shortly :-)

willglynn:
As far as I can tell, the only external difference between this version and the previous is that `readFrom()` now returns the number of bytes read in case of an error, instead of always returning zero. That and, on little-endian platforms, it causes the `io.Reader` to read straight into the `content []uint16` and `bitmap []uint64` members without additional copying.

lemire:
Excellent. In an upcoming commit, I will add your name to the README.

lemire:
https://github.com/tgruben/roaring/commit/fbd7373fa64fe981c817703c500a08aa293d9246

PR #44, opened 2015-11-24
Speed up sequential Adds (for sparse bitmaps)

bpot:
We almost always build bitmaps sequentially and I imagine that's not uncommon. This adds a special case to `(*arrayContainer).add()` which just does a simple append if the value is larger than all others in the container.

```
benchmark                           old ns/op     new ns/op     delta
BenchmarkSequentialAdd-4            38711199      16057751      -58.52%
```

Doesn't seem to hurt random add performance:

```
BenchmarkSetRoaring-4               71.4          70.5          -1.26%
```

### DIFF
--- a/roaring.go
+++ b/roaring.go
@@ -185,7 +185,7 @@ func (rb *RoaringBitmap) Equals(o interface{}) bool {
 // Add the integer x to the bitmap
 func (rb *RoaringBitmap) Add(x uint32) {
 	hb := highbits(x)
-	ra := rb.highlowcontainer
+	ra := &rb.highlowcontainer
 	i := ra.getIndex(hb)
 	if i >= 0 {
 		var c container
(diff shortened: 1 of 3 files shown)


roaring.go
```diff
@@ -185,7 +185,7 @@ func (rb *RoaringBitmap) Equals(o interface{}) bool {
 // Add the integer x to the bitmap
 func (rb *RoaringBitmap) Add(x uint32) {
 	hb := highbits(x)
-	ra := rb.highlowcontainer
+	ra := &rb.highlowcontainer
```
bpot: Take a reference to the highlowcontainer to avoid copying the struct to the local variable. Speeds up things a little.

lemire:
Excellent. The Java version does have a similar optimization: 

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/Util.java#L202-L205

I cannot tell whether the Go version never had this optimization or whether it was mistakenly removed. In any case, it needs to be there, you are right.

Issue #27, opened 2015-04-10
And, Or, Xor and AndNot in-place functions should be optimized


lemire:
Currently, the binary And, Or, Xor and AndNot in-place functions do not attempt to do the computation in-place. 

A genuine in-place implementation, as is available in the Java counterpart https://github.com/lemire/RoaringBitmap would use less memory and be faster in some instances (because it would avoid memory allocation and unnecessary copies).

lemire:
Partially resolved: And and AndNot functions have been optimized.

It remains to optimize Or and Xor similarly.

PR #45, opened 2015-12-03
Partial in place Xor

bpot:
This still builds a new container if there is a container for a given key in both bitmaps. However it is much faster if the second bitmap is much sparser than the first and slightly faster for two random bitmaps with the same cardinality and range.

Fulfills part of #27.

### Benchmark

```
benchmark                           old ns/op     new ns/op     delta
BenchmarkXorLopsided-4              441580        38068         -91.38%
BenchmarkXor-4                      1319948       1204654       -8.73%


benchmark                  old allocs     new allocs     delta
BenchmarkXorLopsided-4     3084           71             -97.70%
BenchmarkXor-4             3084           3059           -0.81%

benchmark                   old bytes     new bytes     delta
BenchmarkXorLopsided-4      313552        36512         -88.36%
BenchmarkXor-4              443680        351968        -20.67%
```

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -466,3 +466,47 @@ func BenchmarkSequentialAdd(b *testing.B) {
 		}
 	}
 }
+
+func BenchmarkXor(b *testing.B) {
+	b.StopTimer()
+	r := rand.New(rand.NewSource(0))
+	s := NewRoaringBitmap()
+	sz := 100000000
+	initsize := 65000
+	for i := 0; i < initsize; i++ {
+		n := uint32(r.Int31n(int32(sz)))
+		s.Add(n)
+	}
+	x2 := NewRoaringBitmap()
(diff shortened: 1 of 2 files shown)


lemire:
@bpot  For extra credits, would you throw in at least one unit test so we know that the function is a bit sane with respect to what it outputs?

bpot:
@lemire I think there are already a couple https://github.com/tgruben/roaring/blob/7747b7aa1bd4e8ed06edd304a59dfe5d67ae1cc9/roaring_test.go#L1066-L1067 https://github.com/tgruben/roaring/blob/7747b7aa1bd4e8ed06edd304a59dfe5d67ae1cc9/roaring_test.go#L1182-L1183

They were failing as I was developing and fixing bugs.

Happy to add more if those aren't sufficient.

lemire:
@bpot  Ok. I'll merge. Sorry about that.

PR #48, opened 2016-01-04
cosmetic changes and simplify FastAnd

brentp:
moving the if statement into the loop removes a level of indentation


lemire:
Great. Will add your name to list of contributors.

Issue #64, opened 2016-09-16
possible Flip() bug?


steveyen:
Hi, thanks for the cool work on roaring bitmaps.

Here's something strange that I originally encountered while trying to write some automated fuzz testing for roaring.

It's possible I'm using the Flip() API incorrectly, or the bug is really in the willf.BitSet that I'm comparing roaring against, but please see here for a simple test...

https://github.com/steveyen/roaring/blob/master/roaring_test.go#L1685
```
		bm := roaring.NewBitmap()
		bs := bitset.New(0)

		bm.AddInt(0)
		bs = bs.Set(0)

		bm.Flip(1, 2)
		bs = bs.Flip(1)

		So(equalsBitSet(bs, bm), ShouldEqual, true) // Fails here.
}
```

lemire:
Yes. Good catch. Fixed in v0.2.8. See unit test I added: https://github.com/RoaringBitmap/roaring/blob/master/roaring_test.go#L12-L32

lemire:
Please keep hunting for bugs!

lemire:
Please considering issuing a PR with your new tests. We never have enough good tests!

PR #65, opened 2016-09-19
fuzz testing via the smat library

steveyen:
One bug found from the fuzz testing runs so far was...

https://github.com/RoaringBitmap/roaring/issues/64

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*~
+roaring-fuzz.zip
+workdir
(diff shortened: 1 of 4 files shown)


lemire:
Though tests fail due to missing dependency, I am confident that this can be fixed later. Merging. Thanks.

lemire:
As expected, it passes travis' tests now. Thanks.

Issue #37, opened 2015-08-18
Implement run containers


lemire:
For better compression when there are long runs of consecutive values, the Java version has run containers (as of version 0.5). They should be implemented :

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RunContainer.java

These can be created by the "runOptimize" function...

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1286

These containers can improve the compression ratio in some cases.

lemire:
This is now used in Apache Spark and Druid with good results.

lemire:
c.c. @brentp

brentp:
so then, for every method, it needs to implement, e.g. 
```
bitmap.intersects(bitmap)
array.intersects(bitmap)
array.intersects(rle)
rle.intersects(bitmap)
```

lemire:
@brentp 

Sadly, yes, it is a pain. On the plus side, there is a complete Java implementation that has been tuned and debugged. So copy-and-paste buys a lot of progress. Believe me, the first implementation (in Java) is surely orders of magnitude more troublesome than a simple port of the idea.

The benefits can be huge however in practice.

brentp:
the benefits are in space? or time? must be space. I guess there could be a method rle.convert() that returns either a bitmap or an array depending on it's own density?

lemire:
There are important gains in both space and time.

Here is the preprint :

http://roaringbitmap.org/RunRoaringBitmap.pdf

It is really very good. I was surprised at how well it did. Better yet: the good results were preserved when it went into production.

brentp:
so the rle is single uint16 array (slice in go) with alternating values of [num_i, runlen_i-1, num_j, runlen_j-1, ...] 
is there an advantage of that over 2 separate arrays?

lemire:
@brentp 

We interleave the data because it is always used together. Otherwise, the code is unnecessarily  complex and it is likely that there are more cache misses. Think about a binary search to check whether a value is contained. With two arrays, each access hits two cache lines. You can almost be sure to have twice as many cache misses.

Separating the data is indeed the correct default strategy when the data is commonly *not* used together. Then you have less data to go through and possibly higher speed. I don't think it is one of these cases.

brentp:
so from the paper, it looks like Roaring without runs is almost always at least as fast as roaring+runs except for entire unions where roraing+runs is much faster.
[...]

PR #70, opened 2016-11-26
RunContainer32 implements a run-length

glycerine:
(interval based) uint32 set. Intersect
and Union are complete and tested.

There is more to do to integrate with roaring.Bitmap,
but this is a nice chunk of functionality. A good start.


glycerine:
TravisCI seems to itself be the cause of some issue. I can't make it exactly what. It doesn't appear to be code related, but something about not finding a repo for this job... 
```
...
PASS

coverage: 86.9% of statements

ok  	github.com/RoaringBitmap/roaring	34.835s

The command "go test -v -covermode=count -coverprofile=coverage.out" exited with 0.

0.98s$ $HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci -repotoken KOlKyOXz0qSjAqvfTF28RzNlr3unxjrLh

Bad response status from coveralls: 422 - {"message":"Couldn't find a repository matching this job.","error":true}

The command "$HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci -repotoken KOlKyOXz0qSjAqvfTF28RzNlr3unxjrLh" exited with 1.

Done. Your build exited with 1.
```

glycerine:
Update - posting here, but this should be read in the context of #37.
as of b4b9700827e9b2041cc95312f6f27a5f4756929a, the #70 PR is almost complete. 

runContainer16 is missing only 3 methods of the container interface:
xor, lazyIOR, and lazyOR.  These shouldn't be difficult, although I'm not exactly clear on
the semantics of the lazy methods.

Also the container interface was somewhat inconsistent, with methods like add and remove sometimes acting inplace and sometimes not; I've tried to address this, for example, by splitting into iadd and iaddReturnMinimized. The convention for the ReturnMinimized suffixed methods is that they may change container type. This simplified certain aspects of testing.

@lemire I may not have time -- at least for a while -- to work on this again in the short term, so I invite anyone is able to finish this up. I wrote tests as I went, so coverage of the central file, rle16.go is very good; currently at 94%.

tests are still needed for the following. I assume that existing arraycontainer_test.go or bitmapcontainer tests can be readily adapted.

* fillLeastSignificant16bits()
* getShortIterator()
* getSizeInBytes() 
* iaddRange()
* iremoveRange()
* not(firstOfRange, lastOfRange int) container
* equals()
* intersects

these three need both -- i.e. both tests and implementation. Implementation is stubbed out in rlei.go:
* lazyIOR()
* lazyOR()
* xor()

lemire:
@glycerine  Excellent. I'll review.

brentp:
@glycerine is there a downside to having the `iremove`, `iadd` methods always return the modified container?
by that I mean to copy, e.g. `iadd` into `iaddReturnMinimized` and rename `iaddReturnMinimized` to just iadd and rely on documentation? 
Or is there a better name, maybe `adds`, `removes` ? I guess it seems a shame to have a larger internal api when it's not used much.

rle.go
```diff
@@ -0,0 +1,1325 @@
...
+// alreadySorted to false, and we will sort them in place for you.
+// (Be aware of this side effect -- it will affect the callers
+// view of vals).
+//
+// If you have a small number of additions to an already
+// big runContainer32, calling Add() may be faster.
+func (rc *runContainer32) set(alreadySorted bool, vals ...uint32) {
+
+	rc2 := newRunContainer32FromVals(alreadySorted, vals...)
+	p("set: rc2 is %s", rc2)
```
brentp: what is `p()`?

rlecommon.go
```diff
@@ -0,0 +1,163 @@
...
+	"fmt"
+	"sort"
+)
+
+// common to rle32.go and rle16.go
+
+// rleVerbose controls whether p() prints show up.
+// The testing package sets this based on
+// testing.Verbose().
+var rleVerbose bool = true
```
brentp: shouldn't this default to false?

glycerine: yes. my bad. left it on while testing.

rle.go
glycerine: rlecommon.go:18
```
// p is a shorthand for fmt.Printf with beginning and                                                                         
// trailing newlines. p() makes it easy                                                                                       
// to add diagnostic print statements.                                                                                        
func p(format string, args ...interface{}) {
    if rleVerbose {
        fmt.Printf("\n"+format+"\n", args...)
    }
}
```
I find godef essential for jumping to definitions. https://github.com/rogpeppe/godef
`(global-set-key [f12] 'godef-jump) ;; in my ~/.emacs`

rle.go
```diff
@@ -0,0 +1,1325 @@
+package roaring
+
+//
+// Copyright (c) 2016 by the roaring authors.
+// Licensed under the Apache License, Version 2.0.
+//
+// We derive a few lines of code from the sort.Search
+// funcion in the golang standard library. That function
```
glycerine: got it. will push a fix shortly.

glycerine:
@brentp 
> is there a downside to having the iremove, iadd methods always return the modified container?

Not sure. I'm open to reverting it.  I just found it confusing that the inplace 'i' prefix convention was violated, so I sought to correct that. But now that I see that there are space optimizations at stake that may not be a worthy design goal. Might be better just to drop the naming convention instead.

glycerine:
I ported the andNot routine from the C to Go, and fixed isubtract so that it is very frugal with memory allocation. Also implemented xor.

There remains for implementation only 

* lazyIOR()
* lazyOR()

could somebody explain what makes them lazy?

lemire:
@glycerine 

I think that if you are only missing this, then 

1. You are pretty much a hero...
2. You are pretty much done...

Lazy functions are optional and can be simply wrapper around non-lazy functions.

The idea of "laziness" is as follows. It is inspired by the concept of lazy evaluation you might be familiar with (functional programming and all that). So a roaring bitmap is such that all its containers are, in some sense, chosen to use as little memory as possible. This is nice. Also, all bitsets are "cardinality aware" so that you can do fast rank/select queries, or query the cardinality of the whole bitmap... very fast, without latency.

However, imagine that you are aggregating 100 bitmaps together. So you OR the first two, then OR that with the third one and so forth. Clearly, *intermediate* bitmaps don't need to be as compressed as possible, right? They can be in a "dirty state". You only need the end result to be in a nice state... which you can achieve by calling ``repairAfterLazy`` at the end.

The Java/C code does something special for the in-place lazy OR runs. The idea is that instead of taking two run containers and generating a new one, we actually try to do the computation in-place through a technique invented by @gssiyankai (pinging him!). What you do is you check whether the host run container has lots of extra capacity. If it does, you move its data at the end of the backing array, and then you write the answer at the beginning. What this trick does is minimize memory allocations.

It is not very complicated, but I am not sure you need to worry about this at all. We could just mark it as a to-do, a possible optimization.

glycerine:
@lemire sounds like lazy would be great to have, but I'll leave it as an optimization for later/someone ese. I added your comments in line to rlei.go and marked the lazyIOR/Or to panic with a TODO message if invoked.

gssiyankai:
@glycerine 
Indeed, the Java version of Roaring has an optimized in-place OR: as explained by @lemire, the less the reallocation, the better. We observed a huge benefit to do so, between 20 and 80%, in terms of memory and runtime, mainly because there is less stress on the garbage collector I believe. I'm not a big expert of Go, but if there is the same kind of memory management and garbage collection as in Java, I think you can expect the same kind of improvements.

glycerine:
@petermattis @knz when cockroachdb gets to optimizing joins, you should definitely checkout roaring and @lemire's research on fast compressed bitmaps. I contributed a Go Run-length-encoding implementation in this PR, as lemire's papers shows this can be a major win. [1][2][3]

[1] http://roaringbitmap.org/RunRoaringBitmap.pdf

[2] https://arxiv.org/pdf/1603.06549.pdf

[3] http://roaringbitmap.org/

lemire:
@glycerine 

Great. It does sound like you are some kind of Golang hero.

lemire:
@RoaringBitmap/go-roaring We have a huge (and possibly very important) PR. We definitively can use all the eyes we got to review this. I will go over it myself, but if other people are interested in having a look, please consider doing so in the coming days.

lemire:
@glycerine 

Ok. Good stuff. Let us see what is missing.


1. I think that ``rangeOfOnes`` should generate a run container, see https://github.com/glycerine/roaring/blob/master/roaringarray.go#L53-L59
 It is trivial to implement.

2. We need something like a ``RunOptimize`` function that looks at the existing containers and convert them to a run container if needed (if it is more economical to represent the data as a run container)... see the C implementation... https://github.com/RoaringBitmap/CRoaring/blob/master/src/roaring.c#L954-L966  and java implementation...  https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1859-L1869
 It is probably not too difficult, but it is not trivial. The main difficult is to count the numbers of "runs" in array and bitset containers.  Counting the number of runs in an array container is not too difficult. For bitset containers, it is less trivial, but we have existing code... see the java version that counts the number of runs in a bitset container... https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/BitmapContainer.java#L875-L892

3. Optionally (this is a performance issue), we should check add a ``isFull`` to containers which checks whether the cardinality is ``1<<16`` (the maximum). When computing a union, we should check whether result ``isFull`` and in these cases, we should convert the result (if it is a bitset container) to a run container... since run containers are much more efficient to represent full containers. (This is a very important optimization in some instances.) I expect that it is not hardl to implement, just a check followed by a conversion if needed.

4. After implement items 1 and 2 (and maybe 3), we need to "upgrade" our unit tests to include calls to ``RunOptimize``. This can be done by copying and pasting some tests in https://github.com/RoaringBitmap/roaring/blob/master/roaring_test.go and adding a version where we call ``RunOptimize`` after creating the bitmaps. That's easy to implement. Same here... https://github.com/glycerine/roaring/blob/master/fastaggregation_test.go We should probably update our ``smat`` tests to include  ``RunOptimize``. https://github.com/RoaringBitmap/roaring/blob/master/smat.go

5. At this point, I think we will find that we are missing some glue... e.g., a function like this one does not currently expect a run container...
https://github.com/glycerine/roaring/blob/master/bitmapcontainer.go#L205-L213 Running the tests, include the  ``smat`` tests (which need to be run more or less manually) might expose some bugs.

6. In due time, we should make that the serialization format is compatible with the Java and C formats. (Many users enjoy cross-language interoperability.) I think we should do this last.

glycerine:
@lemire sounds good. The serialization idea is good (6), but seems like an orthogonal project. I suggest making a separate issue for it if there is not one already.

lemire:
@glycerine Well, I volunteer to fully implement (6)... and yes it can be a separate issue.

glycerine:
https://github.com/RoaringBitmap/roaring/issues/75 should be resolved before merging this pr.

glycerine:
rebased against master, fixed new stat test, squashed into 265b1aaf549aa3f31675d38d7f367e523426cc24

glycerine:
status update: after some implementation, revision and tests were added...

0. these still need tests:

- fillLeastSignificant16bits()
- getShortIterator()
- getSizeInBytes()
- iaddRange()
- iremoveRange()

1. these need both tests and implementation. See notes above for descriptions.

- lazyIOR()
- lazyOR()
- RunOptmize()
- isFull()

glycerine:
Progress on the RunOptimize optimize front ... it is almost there. As of https://github.com/glycerine/roaring/commit/35d8260f446a6741bf98d7cc6efab614e50ad21d 

all container types now implement numberOfRuns()
c.f.
https://github.com/glycerine/roaring/blob/rle/bitmapcontainer.go#L670
https://github.com/glycerine/roaring/blob/rle/arraycontainer.go#L738
https://github.com/glycerine/roaring/blob/rle/rle16.go#L1481

toEfficientBitmap() has been ported and tested. Please review if this is the actual heuristics that are preferred:
https://github.com/glycerine/roaring/blob/rle/rlei.go#L553

glycerine:
update: I implemented a primitve runOptimize() on roaringArray in the latest. See

https://github.com/glycerine/roaring/commit/c1c5d3dfba35b96a3840346dbd47798153c38d2c#diff-61f3efa34dc8855605931ff965532d90R91

I'm not sure how the lazy system works, so I'll need to some input there -- what should happen to the the roaringArray.needsCopyOnWrite and copyOnWrite and keys values when toEfficientContainer() is called in runOptimize() ?

lemire:
@glycerine  

The implementation of ``toEfficientContainer`` looks good.

The unit tests seem to fail due to missing dependencies. As a next step, I'd suggest updating the ``.travis`` file so that we can get back to a working CI...
https://github.com/glycerine/roaring/blob/master/.travis.yml#L3-L8

> I'm not sure if the heuristics are supposed to be symmetric between container types.

I'm not sure what you mean exactly. Definitively, for non-in-place operations with OR, AND and XOR, the operations are symmetrical. For ANDNOT, symmetry is not possible. For the in-place operations, symmetry is not a given, of course.

> I'm not sure how the lazy system works, so I'll need to some input there -- what should happen to the the roaringArray.needsCopyOnWrite and copyOnWrite and keys values when toEfficientContainer() is called in runOptimize() ?

So as not to confuse the terminology, I suggest using the precise term copy-on-write instead of lazy... (though lazy is correct, of course). In the C implementation (https://github.com/RoaringBitmap/CRoaring/blob/master/src/roaring.c#L958-L963) where we also have copy-on-write, we "unshare" the containers before applying run optimization. 

So I think it should go something like this...

```
if runoptimizationisneeded {
	ra.containers[i] = ra.containers[i].runOptimize()
	ra.needCopyOnWrite[i] = false
}
```

This means that if another bitmap also share the same containers, it won't be affected by the fact that you call "runoptimize" on another bitmap.

lemire:
Just so we can keep track, here is where I think we are:

- [x] ``rangeOfOnes`` should generate a run container (Done: https://github.com/glycerine/roaring/pull/1)
- [ ] Implement ``RunOptimize`` (partially done, in progress)
- [ ] Add ``isFull`` method and convert full bitmaps to run containers when they are full
- [ ] Upgrade unit tests so that they call ``RunOptimize``, including the ``smat`` tests
- [ ] Make sure we pass the updated unit tests

lemire:
@brentp 

You have outstanding "change requests" on this PR. Is this still valid?

brentp:
I changed to approved. I haven't looked at anythingn since my initial review, but I think it'd be good to get this in.

glycerine:
This is in good shape. I've squashed it into one commit for convenience.

tests for the runContainer implementation done:
- [x] fillLeastSignificant16bits()
- [x] getShortIterator()
- [x] getSizeInBytes()
- [x] iaddRange()
- [x] iremoveRange()

other checklist items:
- [x] serialization of Bitmaps including the runContainer in both "classic" Roaring (still the default) and the new msgpack2 are done.
- [x] ``rangeOfOnes`` should generate a run container (Done: https://github.com/glycerine/roaring/pull/1)
- [x] Implement ``RunOptimize``
- [x] Add ``isFull`` method; and
- [x]  convert full bitmaps to run containers when they are full
- [ ] Upgrade unit tests so that they call ``RunOptimize``, including the ``smat`` tests
- [ ] Make sure we pass the updated unit tests

things are pretty well tested now. If somebody wants to do smat tests before merging, please go ahead. @lemire, I'm not sure what you had in mind there. But I think it should be fine to merge and figure those out later.

glycerine:
I opened #79, #80, #81 to make notes of additional future TODOs.

lemire:
Noted.

lemire:
@glycerine 

I have issued a PR against your fork, mostly with a little bit of additional testing: 
https://github.com/glycerine/roaring/pull/3

glycerine:
In case github doesn't make it obvious, I've pushed all merged pull-requests from glycerine/roaring to this branch (glycerine/rle) as well.

glycerine:
With the addition of the 067 test (TestAllContainerMethodsAllContainerTypesWithData067 in rlei_test.go), all containers types must agree with each others results over the same operations -- on random data, and over 100 passes of randomized pairwise container operations.

lemire:
I am restarting a failed CI.

lemire:
@glycerine  Ok. Do you recommend immediate merger?

glycerine:
After fuzzying the top end of the range (near MaxUint16), I found and fixed some additional overflow issues.

I now recommend merging it.

lemire:
Merged. This is a massive piece of work.

glycerine:
Seriously!

PR #73, opened 2016-12-06
Bitmap.Stats() method proposal

maciej:
Hi,

Here is a proposal of a `Stats()` method, similar to one in [gocroaring][goc-stats].

It gathers the same statistics, however unlike the _gocroaring_ version it returns a struct instead of a `map[string]uint64`. I think there are two advantages of this approach:
* discoverability of the gathered statistics when using an editor with auto-complete
* a mistyped statistic name in the client application will result in a compile time error instead of a runtime one

I'm hoping for comments on this PR. I'll add missing tests once the method interface is agreed on. :-)

[goc-stats]: https://github.com/RoaringBitmap/gocroaring/blob/master/gocroaring.go#L239

### DIFF
--- a/roaring.go
+++ b/roaring.go
@@ -1111,3 +1111,36 @@ func (rb *Bitmap) GetCopyOnWrite() (val bool) {
 func FlipInt(bm *Bitmap, rangeStart, rangeEnd int) *Bitmap {
 	return Flip(bm, uint64(rangeStart), uint64(rangeEnd))
 }
+
+type Statistics struct {
+	Cardinality           uint64
+	Containers            uint64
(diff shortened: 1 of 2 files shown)


lemire:
This seems perfectly fine to me. In fact, you make a good point that gocroaring should be modified.

lemire:
I don't expect that there will be bugs, but I do appreciate that we should write at least one unit test to make sure it is sane. Will merge as soon as you say "go".

maciej:
@lemire great! I'll write the test cases and get back to you!

maciej:
@lemire I've added 3 basic test cases. I think they cover the method well enough.

PR #77, opened 2016-12-13
runContainer16/32 and msgpack2/snappy serialization

### LINKED ISSUE #37: Implement run containers

### LINKED ISSUE #76: thoughts on serialization formats
Thoughts on serialization... things that could save a bunch of time...

When I need quick and portable serialization, I typically use msgpack2 from http://msgpack.org. It's a binary version of JSON that has a readable spec and is portable to all languages.

A reason I like msgpack2: there is an easy to use library for Go that is very fast.
Fast to deploy and fast when reading/writing.

https://github.com/tinylib/msgp in particular will parse your .go files and do codegen for your Go structs, making setting up serialization a 30 second task. All that is required is that your struct fields be Exported (capitalized).

Usually compression isn't needed afterwards, depending on the entropy of your data. You have to measure, but sometimes it is a win. If desired, then snappy can be easily applied without sacrificing decoding speed. Other general purpose compressors like gzip and bzip will sometimes compress slightly more, but are super slow to decode, hogging cpu and wasting valuable time at runtime. I usually use my implementation of the streaming snappy format here, https://github.com/glycerine/go-unsnap-stream, as it comes with nice command line tools (snap and unsnap).  Since Google open sourced snapped (http://google.github.io/snappy/) quite a while ago, there are libraries for most languages for this as well.

Finally, one other nice thing about msgpack2 is that you don't need to predefine a schema; like JSON the data is self describing. It gives you some data-evolution capability (based of field names rather than field numbers) while being dynamic language friendly. There is no need for a pre-compiler (as in protobufs/thrift/capnproto) and java and python can read msgpack2 with the available libraries. msgpack2 is quite a bit more strongly typed that JSON however, and you can quickly move binary data without a whole lot of base64 encoding fuss.

That said, if one really wants a schema for future proofing the data evolution, then protobufs are also highly tuned and have libraries for the big 4 (Go, C++, Java, python).  Also I used to maintain a capnproto implementation for Go, so I can discuss that if it is of interest. However the java bindings were never finished, so I didn't really consider capnproto a choice here where java support is expected.

glycerine:
https://github.com/RoaringBitmap/roaring/pull/70 had the problem that the tests wouldn't all pass because the serialization wasn't done for runContainer16. In this `serialization_msgp` branch I finished it off using the serialization methods (msgpack2 for serialization and snappy for compression) discussed in https://github.com/RoaringBitmap/roaring/issues/76.

Fixes #37
Fixes #76 

All tests are green. From the checklist on #70:

- [x] serialization of Bitmaps including the runContainer is done in this PR.
- [x] ``rangeOfOnes`` should generate a run container (Done: https://github.com/glycerine/roaring/pull/1)
- [x] Implement ``RunOptimize``
- [x] Add ``isFull`` method; and
- [ ]  convert full bitmaps to run containers when they are full
- [ ] Upgrade unit tests so that they call ``RunOptimize``, including the ``smat`` tests
- [ ] Make sure we pass the updated unit tests

these routines still need tests for the runContainer implementation
- [ ] fillLeastSignificant16bits()
- [ ] getShortIterator()
- [ ] getSizeInBytes()
- [ ] iaddRange()
- [ ] iremoveRange()

I open this as a distinct branch in case nobody wants the msgpack2/snappy serialization.

Of course I think it's great. But if there are back-wards compatibility concerns then it may be a non-starter.


glycerine:
If the serialization format is acceptable, I suggest we merge this and make the remaining checklist items into new Issues. I don't want to maintain this big a change on a branch for very long.

lemire:
Got your message. I may not have time to look at this immediately. Here is my feeling regarding serialization. There is no problem in supporting several formats. However all roaring libraries so far support a designed binary format that enables cool things like memory-mapped bitmaps... and it really easy to implement so there is no reason not to implement it, really. However, we can and probably should support more than one serialization format.

Performance should be assessed.

I'll have a look. Soon.

glycerine:
> enables cool things like memory-mapped bitmaps...
> Performance should be assessed.

Agreed. Perhaps surprisingly, I found on one past application that using msgpack2/snappy was faster than memory-mapping, because you do much less disk I/O. Fewer bytes that have to be read from disk == faster.

lemire:
This is great, and I think we should merge this ASAP.

There is one issue to be resolved. I think we want to keep the existing ``writeTo`` and ``readFrom`` methods.  We have an efficient  format that is supported across a wide range of libraries, and is being used in important production systems, and whose performance and data size has been documented and well tested. We have that ``gocroaring`` (https://github.com/RoaringBitmap/gocroaring), ``RoaringBitmap`` (https://github.com/RoaringBitmap/RoaringBitmap), ``CRoaring`` (https://github.com/RoaringBitmap/CRoaring), ``CRoaring.Net`` (https://github.com/RogueException/CRoaring.Net), ``croaring-rs`` (https://github.com/saulius/croaring-rs) and ``PyRoaringBitMap`` (https://github.com/Ezibenroc/PyRoaringBitMap) all have the same serialization format.

So  I suggest we keep ``writeTo``, ``serializedSizeInBytes``, ``GetSerializedSizeInBytes`` and ``readFrom`` methods as they are, and add `writeToMessagePack``, ``serializedMessagePackSizeInBytes``, ``GetSerializedMessagePackSizeInBytes`` and ``readFromMessagePack`` if you feel  that MessagePack is an important contribution.

If adding support to run containers to ``writeTo`` and ``readFrom`` is an issue,  I will handle it, it is not very difficult.

lemire:
@glycerine (Reposting here) The format specification can be found there: 

https://github.com/RoaringBitmap/RoaringFormatSpec

lemire:
@glycerine  BTW you have now two distinct PRs. I understand that they correspond to different branches, but once everything is sorted out, it would be nicer if you'd be explicit regarding what must be merged when if there is any dependencies. Or, short of that, you could cancel one of your PRs, and put everything into one unified PR. I have no preference, but I'd just want to avoid unnecessary messy mergers.

glycerine:
@lemire sounds good. Thanks for the format docs. I'll try to implement the format. Is there a place where I could extend it to allow runContainer32 as well?  That was one of my long term interests--since I'll always be starting with a [0,2^24-1] container -- for my application this is a super common case.

I should be able to get to it this weekend. I'll consolidate the pull requests into one.

lemire:
@glycerine 

I think there is end to the different variations one can add... but it is nice to have interoperability when needed.

glycerine:
closing in favor of updates to #70.

PR #83, opened 2016-12-18
Fix deserialization of empty bitmaps

bpot:
Cast size to an uint16 so it overflows correctly in the case where there
are 0 containers.

Without this change trying to deserialize an empty bitmap returns an error: `error in readFrom: could not read the runContainer bit flags of length 8192 bytes: EOF`

I'm not sure if this is the best fix, it seems like an empty bitmap is a bit of a grey area -- based on reading the format spec it sounds like there should always at least be 1 container.


lemire:
I'll merge this since it appears harmless, but there is a deeper issue. I think that our serialization code generates the wrong cookie header in some instances. I'll see about fixing this quickly.

lemire:
Issue should be fixed in 0.3.1.

PR #84, opened 2016-12-20
some perf tuning

glycerine:
faster newRunContainer32FromBitmapContainer; countTrailingZeros is done in assembly, replacing numberOfTrailingZeros.

there's more too do, but this is worthwhile to merge; it speeds up the newRunContainer constructor by 4x.


arraycontainer.go
```diff
@@ -439,6 +439,7 @@ func (ac *arrayContainer) lazyorArray(value2 *arrayContainer) container {
 }
 
 func (ac *arrayContainer) and(a container) container {
+	//p("ac.and() called")
```
lemire: comments should probably be removed before merging the PR.

lemire:
Ok. Good stuff. I am just worried about the assembly call here. Though you'd think it would be faster, we really should gather data.

lemire:
@glycerine 

I have run a small improvised benchmark which seems to indicated that the assembly approach is 2x slower (this is *bad*):

```
$ go test -bench=.
testing: warning: no tests to run
BenchmarkDeBruijn64-4   	 1000000	      1644 ns/op
BenchmarkASM-4          	  500000	      2531 ns/op
PASS
ok  	_/Users/lemire/tmp/ctz	3.746s
```

Here is the gist of my benchmark...

```go
func BenchmarkASM(b *testing.B) {
      b.StopTimer()
    r:=1000
    array := make([]uint64, r, r)
    for i := 0; i < int(r); i++ {
        array[i] = uint64(i)
    }
    b.StartTimer()
    sum := 0
    for j := 0; j < b.N; j++ {
        for i := r; i > 0; i-- {
            sum += countTrailingZeros(array[r-1])
        }
    }
}
```

arraycontainer.go
glycerine: Actually I meant to discuss this when the opportunity arose.

Since print based debugging is the reality for all Go programmers, I actually prefer to leave the commented out prints in place. Why make someone go through the hassle of carefully retyping all that again just to debug something the next time. Often that person is me! The comments don't hurt anything, and they do make it much easier to track down problems.

glycerine:
> seems to indicated that the assembly approach is 2x slower

That's interesting and surprising indeed. 

I had thought about including the benchmarks I did, but then that would have meant keeping both new and old code available, which kind of missed the point/seemed off. But I'll recreate them.

arraycontainer.go
lemire: @glycerine 

I am sympathetic to this point of view... 

See http://lemire.me/blog/2016/06/21/i-do-not-use-a-debugger/ 

But I have concerns...

1. Leaving printouts does reduce readability. This seems undeniable.
2. Is this an established best practice? If it were, then we could, at least, rely on a standard for how such printouts should appear...

At a minimum, I think we need to document these printouts. This code gets modified by lots of different people.

glycerine: In fact I already felt the need to maintain a separate fork because the comments got stripped out of other parts. It really makes maintenance so much easier to have them still available.

lemire:
@glycerine 

I'm not sure we need comparison benchmarks within the code, but I think it would be useful to have before/after benchmarks as part of the PR so that we can quantify progress and avoid setbacks.

> That's interesting and surprising indeed.

Assembly calls are not inlined in Go so you have a significant function call overhead to contend with. I think that there is a hidden defer as well.

The deBruijn approach can be as low as one instruction (for ``x & -x`` using BMI) followed by one multiplication (3 cycles latency by high throughput) and one shift... then one load. So it is not going to be terrible.

arraycontainer.go
glycerine: > At a minimum, I think we need to document these printouts. This code gets modified by lots of different people.

I don't know what this would mean. They seem pretty self-evident. What do you have in mind?

lemire: @glycerine 

> I don't know what this would mean. They seem pretty self-evident.
 
I am sure you agree that coding is also very much a social activity where esthetic matters. For example, the Go language went with a very clean look.

I am not sure that these printouts are pretty. 

Now, there is an easy workaround. Just adopt a comment convention that is pretty and readable... then have a script turn these comments into printouts when you need them. Of course, this involves abiding by some standard way of writing these comments so that the script can do its work.

I don't think it is hard to do and I hope you'll agree that it is a reasonable compromise that is best on the long run. 

So instead of having...

```
	//p("ac.and() called")
```

We could have...

```
	/// ac.and() called
 ```
... or something... and then a script could turn it (temporarily) into 

```
	fmt.Println("ac.and() called")
 ```

See my point? 

I am sure we could improve on my proposal... but the idea is to keep the source code as pretty as possible, while still allowing you to have your printouts when you want to have them.

glycerine:
Latest tip of this branch, b7f8315e18aa243c0a65cbb24f2ff1d9ef5355bc, provides benchmarks and fixes the other issues raised (except prints; have to handle that separately). @lemire I'm curious about your preliminary negative result -- if you run these benchmarks -- based on a random sampling rather than sequential -- on your machine, do you still see a lack of improvement for the Asm over the deBruinj? I see an 18% improvement with the Asm.

~~~
go test -bench=100 -run -
Benchmark100OrigNumberOfTrailingZeros-4        50000         38803 ns/op
Benchmark100CountTrailingZerosDeBruinj-4      100000         17144 ns/op
Benchmark100CountTrailingZerosAsm-4           100000         14084 ns/op
PASS
ok      github.com/RoaringBitmap/roaring    5.781s
~~~

arraycontainer.go
glycerine: I do like the kernel of this idea; making it easier to shift back and forth.

In practice, I see issues like:

1. we don't have a clean bijection. Inverting the transformation is difficult, because you don't know which fmt.Printf statements should now be commented, and which should not because they were never comments to begin with. One could try to surmount this by decorating the tail of the statement with another comment, but how to pick a comment that won't also have false positives?
2. One frequently wants to uncomment a single/small number of prints, and leave the rest alone. The conversion tool would be a blunt all or nothing per file hammer. I suppose this could be handled by an editor macro, but... starting to sound like lots of work to write a tool and editor macros. Compared to the simplicity of //p this seems crazy.
3. People would have to be educated about the conversion tool as well; this is big
barrier to entry. Any tooling has a high bar to meet compared to the simplicity of "position the cursor on the comment and hit the delete key twice", which is pretty darn easy way to activate a logging statement.

lemire:
@glycerine 

Using your code and some glue, I have documented a benchmark...
http://lemire.me/blog/2016/12/21/performance-overhead-when-calling-assembly-from-go/

Here is the code...

https://github.com/lemire/Code-used-on-Daniel-Lemire-s-blog/tree/master/2016/12/21

My Go code is somewhat simpler than yours however...

```go
func noasm_countTrailingZeros(x uint64) int {
	// x & -x leaves only the right-most bit set in the word. Let k be the
	// index of that bit. Since only a single bit is set, the value is two
	// to the power of k. Multiplying by a power of two is equivalent to
	// left shifting, in this case by k bits. The de Bruijn constant is
	// such that all six bit, consecutive substrings are distinct.
	// Therefore, if we have a left shifted version of this constant we can
	// find by how many bits it was shifted by looking at which six bit
	// substring ended up at the top of the word.
	// (Knuth, volume 4, section 7.3.1)
	return int(noasm_deBruijn64Lookup[((x&-x)*(noasm_deBruijn64))>>58])
}
```

Timings:
```
$ go test -bench=.
testing: warning: no tests to run
Benchmark100CountTrailingZerosDeBruijn-2        1000000          1045 ns/op
Benchmark100CountTrailingZerosAsm-2             1000000          2379 ns/op
PASS
ok     
```

I think it is a demonstration that the assembly approach is not helpful and should not be used.

lemire:
@glycerine  To compare our results, note that I use arrays of size 1000 as opposed to 6000. There seems no reason to risk getting into L2 cache.

If I multiply by numbers by 6, I get 6270 and 14274 whereas you have 17144 and  14084. This suggests to me that the de Bruijn code you are testing is slower than it should be.

Well. You have a conditional clause branching on the word size along with a panic. This seems unnecessary (we process 64-bit words irrespective of the machine word size) and may harm performance. I'm not sure Go will inline your function. I have not investigated further.

glycerine:
good stuff. I reproduced your benchmark conclusions; and caught a bug in the implementation while doing so; you probably want to update the code linked from your blog to contain this bug fix:
~~~
func noasm_countTrailingZeros(x uint64) int {
	// x & -x leaves only the right-most bit set in the word. Let k be the
	// index of that bit. Since only a single bit is set, the value is two
	// to the power of k. Multiplying by a power of two is equivalent to
	// left shifting, in this case by k bits. The de Bruijn constant is
	// such that all six bit, consecutive substrings are distinct.
	// Therefore, if we have a left shifted version of this constant we can
	// find by how many bits it was shifted by looking at which six bit
	// substring ended up at the top of the word.
	// (Knuth, volume 4, section 7.3.1)
        if x == 0 {
            // We have to special case 0; the fomula                                           
            // below doesn't work for 0.                                                       
            return 64
        }
	return int(noasm_deBruijn64Lookup[((x&-x)*(noasm_deBruijn64))>>58])
}
~~~

latest update on this branch/pr uses the deBruijn. I've left the assembly available for ease of benchmark repro-- in case this varies across precise architectures; or if there's hardware engineer who points out this could be optimized further with assembly.

lemire:
@glycerine  We seem to have failing unit tests...

lemire:
@glycerine 

If you were to use TZCNT, the result would be undefined on a zero input.

glycerine:
tzcnt returns 64 on a zero input, as I expect; the test https://github.com/glycerine/roaring/blob/perftune/ctz_test.go#L147 catches this.

Tests all pass on my laptop. I don't understand the travis unit test failure, because it claims to be in a file that I don't have; roaringcow_test.go

~~~
  * /home/travis/gopath/src/github.com/RoaringBitmap/roaring/roaringcow_test.go 
  Line 1850:
  Expected: 'roaring.Statistics{Cardinality:60000, Containers:1, ArrayContainers:0, ArrayContainerBytes:0, ArrayContainerValues:0, BitmapContainers:0, BitmapContainerBytes:0, BitmapContainerValues:0, RunContainers:1, RunContainerBytes:4, RunContainerValues:60000}'
  Actual:   'roaring.Statistics{Cardinality:60000, Containers:1, ArrayContainers:0, ArrayContainerBytes:0, ArrayContainerValues:0, BitmapContainers:0, BitmapContainerBytes:0, BitmapContainerValues:0, RunContainers:1, RunContainerBytes:52, RunContainerValues:60000}'
  (Should resemble)!
~~~

lemire:
@glycerine 

Not that it will matter for us, since inlining should zap the additional check, but if you add the check for zero, you slow things down by 10%: 

```
$ go test -bench=.
testing: warning: no tests to run
Benchmark100CountTrailingZerosDeBruijnCheck-2   	 1000000	      1192 ns/op
Benchmark100CountTrailingZerosDeBruijn-2        	 1000000	      1042 ns/op
Benchmark100CountTrailingZerosAsm-2             	 1000000	      2365 ns/op
PASS
ok  	_/home/dlemire/CVS/github/Code-used-on-Daniel-Lemire-s-blog/2016/12/21	4.650s
```

(To be clear, your code is fine because, as I just wrote, I expect that it will not add any overhead in practice due to compiler optimizations.)

Intel spec:

> LZCNT is an extension of the BSR instruction. The key difference between LZCNT and BSR is that LZCNT provides operand size as output when source operand is zero, while in the case of BSR instruction, if source operand is zero, the content of destination operand are undefined. On processors that do not support LZCNT, the instruction byte encoding is executed as BSR.

So you are right regarding LZCNT if one carefully checks the CPU type.

lemire:
@glycerine  Tests on ``master`` all pass currently.

https://travis-ci.org/RoaringBitmap/roaring/branches

I added a few extra tests to improve our coverage.

glycerine:
yeah, I just needed to rebase. I didn't notice you had made commits to master. fixed now.

lemire:
@glycerine 

We still have a problem:

```
Failures:
  * /home/travis/gopath/src/github.com/RoaringBitmap/roaring/ctz_test.go 
  Line 14:
  Expected: '64'
  Actual:   '140448050458120'
  (Should be equal)
155 total assertions
--- FAIL: TestCountTrailingZeros072 (0.00s)
```

glycerine:
lol.
just as you were pointing out. it's undefined on older arch.

lemire:
LGTM.

lemire:
After some cleaning, I have released v0.3.3.

Issue #85, opened 2016-12-20
Int overflow?


0x4139:
doesn't this overflow int?
https://github.com/RoaringBitmap/roaring/blob/master/util.go#L283

0x4139:
i think it should be `rand.Intn((1 << 30)-1)`

lemire:
@0x4139 
 
I think that @glycerine  assumed that we are working on a 64-bit system. (???)

0x4139:
@lemire yep

0x4139:
@glycerine can you explain why `1 << 29` and not (1<<31)-1 being the largest number on 32 bit systems?

lemire:
@0x4139  @glycerine I'm pretty sure that the answer is ``(1<<31)-1`` though it is maybe better expressed as ``0x7fffffff`` (matter of taste).

glycerine:
@0x4139 It doesn't matter for correctness, and the point is to look at it and quickly know that it doesn't overflow int even on 32-bit systems.

PR #87, opened 2016-12-20
fixes the int overflow

0x4139:
regarding the issue #85

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *~
 roaring-fuzz.zip
 workdir
+.idea
--- a/roaringarray.go
+++ b/roaringarray.go
@@ -8,6 +8,8 @@ import (
 
 	snappy "github.com/glycerine/go-unsnap-stream"
 	"github.com/tinylib/msgp/msgp"
+	"reflect"
+	"unsafe"
 )
 
 //go:generate msgp -unexported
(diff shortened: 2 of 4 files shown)


lemire:
@0x4139  

I am concerned with your commit "moves the utility functions away from the file with conditional builds". 

https://github.com/RoaringBitmap/roaring/pull/87/commits/f341b887fd0371f395ee8c8e9104313bdf7d5516

I don't think it is correct because the Roaring format requires all words to be written using little endian encoding...

https://github.com/RoaringBitmap/RoaringFormatSpec#general-layout

The proper fix here is that uint64SliceAsByteSlice and uint16SliceAsByteSlice should be "file private" to https://github.com/RoaringBitmap/roaring/blob/master/serialization_littleendian.go and never called elsewhere.

c.c. @glycerine

0x4139:
@lemire you might be right, it's my first time i am seing the code, but doing so we need to make some other module

lemire:
@0x4139  For now, I just suggest you remove this commit from the PR, so we can address the issue separately.

Also, please add your name to https://github.com/RoaringBitmap/roaring/blob/master/AUTHORS

lemire:
@0x4139  I don't think we need a new module, we just need to change @glycerine 's recent changes so that we avoid the calls to ``uint64SliceAsByteSlice`` and ``uint16SliceAsByteSlice``. They should not have been allowed in... but I missed it. My fault. Thankfully, it is not hard to fix.

lemire:
To be clear, you should add your name to https://github.com/RoaringBitmap/roaring/blob/master/CONTRIBUTORS

lemire:
@0x4139  If the PR is still relevant, please rebase with master.

glycerine:
this was addressed in #84; you can close without merging.

lemire:
@0x4139  Ok. I am going to close this. Reopen a new PR if needed.

PR #90, opened 2016-12-23
atg. elminate ctz asm. fixes #89

### LINKED ISSUE #89: Build fails for ARM architecture (maybe anything other than amd64?)
I don't know if this an architecture you care about targeting, but it's one my company has been using, and it looks like #84 broke building there. (We use `gomobile` to build a library that we use on iOS.)

Here are the errors I get running `GOARCH=arm64 go build` in my project that uses this library:

```
# github.com/figly/datasearch/vendor/github.com/RoaringBitmap/roaring
../vendor/github.com/RoaringBitmap/roaring/ctz_generic.go:52: undefined: deBruijn64Lookup
../vendor/github.com/RoaringBitmap/roaring/ctz_generic.go:52: undefined: deBruijn64
../vendor/github.com/RoaringBitmap/roaring/serialization_generic.go:56: undefined: buf
```

The last error looks like a pretty simple variable name typo, but I'm not sure what the right fix is for the first two.

glycerine:
Simplify the countTrailingZeros stuff by eliminating the unused assembly.

I thought somebody more skilled at assembly than myself might take a look and
speed up the assembly version. But for now it's not used and makes it more of a hassle to get builds
on different platforms correct, so I'm just deleting it.

Fixes #89.


lemire:
LGTM.

PR #91, opened 2016-12-23
confirm build on arm via cross-compile

glycerine:
here's a small patch to the travis config to tell go to cross-compile for arm as well; this would have caught things like #89.


lemire:
Fantastic.

PR #107, opened 2017-10-17
FastOr: force using bitmap containers

maciej:
This change converts containers in `lazyOR` methods to bitmap containers when computing an intersection.

It follows a similar pattern found in Java's RoaringBitmap [RoaringBitmap.naivelazyor](https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1628) method.

In our production case when using very wide unions this change improves performance significantly.

When evaluating an expression which consists of two wide unions (one of 161 bitmaps and another one of 144) the speed improvements was of a factor of x50.

Before:
```
and:497.120619ms, or:38.185436472s, andNot:12.385317ms
```

After:
```
and:30.566886ms, or:623.676526ms, andNot:10.220867ms
```

Sadly I cannot share the input bitmaps.

### DIFF
--- a/fastaggregation.go
+++ b/fastaggregation.go
@@ -32,8 +32,17 @@ main:
 				}
 				s2 = x2.highlowcontainer.getKeyAtIndex(pos2)
 			} else {
+				c1 := x1.highlowcontainer.getContainerAtIndex(pos1)
+				switch t := c1.(type) {
+				case *arrayContainer:
+					c1 = t.toBitmapContainer()
+				case *runContainer16:
+					if !t.isFull() {
+						c1 = t.toBitmapContainer()
+					}
+				}
 
-				answer.highlowcontainer.appendContainer(s1, x1.highlowcontainer.getContainerAtIndex(pos1).lazyOR(x2.highlowcontainer.getContainerAtIndex(pos2)), false)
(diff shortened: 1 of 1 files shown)


lemire:
I’ll review soon.

lemire:
This code looks very similar to both the Java and C code that we have thoroughly benchmark. Furthermore, you indicate that you are getting good results on private benchmarks. That's good enough for me.

lemire:
I'll issue a release.

maciej:
@lemire great, thank you!

lemire:
PR https://github.com/RoaringBitmap/roaring/pull/331 is going to revert this PR.

Further investigations reveals that it could lead to containers containing mere dozens of entries... Spending 8kB on a few entries can boost performance for sure, but there is a significant tradeoff involved.

In the Java or C++ implementation of roaring, we do not ever go for bitmap containers with less than or equal to 4096 entries. It is a constant. I think it is best to stay with the standard format and not to trade memory usage away for performance. The results could be unfortunate.

Further PRs to improve the performance are certainly invited. I have added some checks to help us avoid such deviations in the future.

PR #109, opened 2017-10-20
Parallel roaring bitmaps

### LINKED ISSUE #38: Parallelize
Roaring is ideally suited for parallel execution, as all containers can be operated upon in distinct threads.

maciej:
In this PR I parallelise roaring `And` and `Or` operations.

I depart from adding any shared aggregators, like in #104. The interface is kept simple – only two methods are added. `ParAnd` and `ParOr` take after `FastAnd` and `FastOr`. Resources (goroutines and channels) are not shared between invocations.

I use a heap as means of iterating through the passed bitmaps. Since t's slightly different from the one in `priorityqueue.go` it's implemented from scratch and not reused.

Tests are copied from `fast aggregation_test.go` and adapted.

I consider adding versions of `ParAnd` and `ParOr` in the future that would use shared resources (similar to ParAggregator from #104).

Fixes #38.

### DIFF
--- a/parallel.go
+++ b/parallel.go
@@ -177,6 +177,13 @@ func appenderRoutine(bitmapChan chan<- *Bitmap, resultChan <-chan keyedContainer
 }
 
 func ParOr(bitmaps ...*Bitmap) *Bitmap {
+	bitmapCount := len(bitmaps)
+	if bitmapCount == 0 {
+		return NewBitmap()
+	} else if bitmapCount == 1 {
+		return bitmaps[0].Clone()
+	}
+
 	h := newBitmapContainerHeap(bitmaps...)
 
 	bitmapChan := make(chan *Bitmap)
@@ -235,6 +242,11 @@ func ParOr(bitmaps ...*Bitmap) *Bitmap {
 
 func ParAnd(bitmaps ...*Bitmap) *Bitmap {
 	bitmapCount := len(bitmaps)
+	if bitmapCount == 0 {
+		return NewBitmap()
+	} else if bitmapCount == 1 {
+		return bitmaps[0].Clone()
+	}
 
 	h := newBitmapContainerHeap(bitmaps...)
 
--- a/parallel.go
+++ b/parallel.go
@@ -1,0 +1,284 @@
+package roaring
+
+import (
+	"container/heap"
+	"runtime"
+)
+
+var defaultWorkerCount = runtime.NumCPU()
+
+type bitmapContainerKey struct {
+	bitmap    *Bitmap
+	container container
+	key       uint16
+	idx       int
+}
+
+type multipleContainers struct {
+	key        uint16
+	containers []container
+	idx        int
+}
+
+type keyedContainer struct {
+	key       uint16
+	container container
+	idx       int
+}
+
+type bitmapContainerHeap []bitmapContainerKey
+
+func (h bitmapContainerHeap) Len() int           { return len(h) }
+func (h bitmapContainerHeap) Less(i, j int) bool { return h[i].key < h[j].key }
+func (h bitmapContainerHeap) Swap(i, j int)      { h[i], h[j] = h[j], h[i] }
+
+func (h *bitmapContainerHeap) Push(x interface{}) {
+	// Push and Pop use pointer receivers because they modify the slice's length,
+	// not just its contents.
+	*h = append(*h, x.(bitmapContainerKey))
+}
+
+func (h *bitmapContainerHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
+	return x
+}
+
+func (h bitmapContainerHeap) Peek() bitmapContainerKey {
+	return h[0]
+}
+
+func (h *bitmapContainerHeap) PopIncrementing() bitmapContainerKey {
+	k := h.Peek()
+
+	newIdx := k.idx + 1
+	if newIdx < k.bitmap.highlowcontainer.size() {
+		newKey := bitmapContainerKey{
+			k.bitmap,
+			k.bitmap.highlowcontainer.getWritableContainerAtIndex(newIdx),
+			k.bitmap.highlowcontainer.keys[newIdx],
+			newIdx,
+		}
+		(*h)[0] = newKey
+		heap.Fix(h, 0)
+	} else {
+		heap.Pop(h)
+	}
+	return k
+}
+
+func (h *bitmapContainerHeap) PopNextContainers() multipleContainers {
+	if h.Len() == 0 {
+		return multipleContainers{}
+	}
+
+	containers := make([]container, 0, 4)
+	bk := h.PopIncrementing()
+	containers = append(containers, bk.container)
+	key := bk.key
+
+	for h.Len() > 0 && key == h.Peek().key {
+		bk = h.PopIncrementing()
+		containers = append(containers, bk.container)
+	}
+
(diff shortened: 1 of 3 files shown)


parallel.go
```diff
@@ -0,0 +1,284 @@
...
+	h := newBitmapContainerHeap(bitmaps...)
+
+	bitmapChan := make(chan *Bitmap)
+	inputChan := make(chan multipleContainers, 128)
+	resultChan := make(chan keyedContainer, 32)
+	expectedKeysChan := make(chan int)
+
+	orFunc := func() {
+		for input := range inputChan {
+			c := toBitmapContainer(input.containers[0]).lazyOR(input.containers[1])
```
lemire: This seems to assume that input.containers.length >= 2. Is this enforced/documented?

parallel.go
```diff
@@ -0,0 +1,284 @@
...
+	h := newBitmapContainerHeap(bitmaps...)
+
+	bitmapChan := make(chan *Bitmap)
+	inputChan := make(chan multipleContainers, 128)
+	resultChan := make(chan keyedContainer, 32)
+	expectedKeysChan := make(chan int)
+
+	andFunc := func() {
+		for input := range inputChan {
+			c := input.containers[0].and(input.containers[1])
```
lemire: This seems to assume that input.containers.length >= 2. Is this enforced/documented?

parallel.go
```diff
@@ -0,0 +1,284 @@
...
+	go appenderRoutine(bitmapChan, resultChan, expectedKeysChan)
+
+	for i := 0; i < defaultWorkerCount; i++ {
+		go andFunc()
+	}
+
+	idx := 0
+	for h.Len() > 0 {
+		ck := h.PopNextContainers()
+		if len(ck.containers) == bitmapCount {
```
lemire: So you compute the intersection, no matter how many containers are involved, and then you discard the computed result if the number of containers wasn't maximal. This seems wasteful. Or do I misunderstand?

parallel_benchmark_test.go
```diff
@@ -0,0 +1,57 @@
+package roaring
+
+import (
+	"math/rand"
+	"testing"
+)
+
+func BenchmarkIntersectionLargeParallel(b *testing.B) {
```
lemire: I'm impressed that there are any benefits whatsoever to parallelize the intersection. That's great. But the reason I am impressed is that intersections are not what kills you, typically... the performance troubles are almost always with unions, those are tremendously expensive at times.  So, I wonder, why did you write a benchmark for intersections, but not for unions (ParOr)?

parallel.go
maciej: Yes, that's correct. If count is one then we can pass the resulting bitmap straight to the _appender_ goroutine: as seen around [line 211](https://github.com/RoaringBitmap/roaring/pull/109/files#diff-283cb5309c30ff97b95802636f247453R211).

parallel.go
maciej: Similar as in the `or` case.

parallel.go
maciej: I actually don't compute the intersection unless a container is available for a given key across all input bitmaps. Computing the intersection would make no sense in such case as the result would always be an empty set / container.
The above code sends the containers to the `andFunc` goroutine only if such containers (for all input bitmaps) are present.

parallel.go
lemire: Can you add a test where you do ParOr(rb) where rb is a single bitmap, just so that things are clear?

maciej: That's very good point. The code actually failed for a one-bitmap case in `ParAnd`. It's now fixed.

parallel_benchmark_test.go
maciej: In the private benchmark I referred to in #107 the improvement is clearly visible:

Par:
```
and:24.346914ms, or:106.765985ms, andNot:7.937437ms
```

Fast:
```
and:49.893867ms, or:448.947547ms, andNot:7.49725ms
```

running on a quad-core 2013' MacBook Pro, Go 1.9.1.

I'm thinking of writing code that would generate similar benchmark data, however I'd prefer to open a separate ticket for that and have it done later if that's OK with you.

lemire: Ok.

maciej:
Regarding the API, I'm starting to have second thoughts about the `Par*` method signatures. Although I like to avoid burdening the users with hard to tune parameters (like channel sizes), I think we should expose the parallelism level of the method, i.e. the number of concurrent goroutines executing `or`s or `and`s. I'd like to change that before we merge.

lemire:
Ok. Anyhow I still have doubts, not about how correct this is, but I don't think that the intersection code does what you think it does.

maciej:
@lemire I think the intersection code is correct. Let me try to explain my thinking:

Let's say we take a bunch of bitmaps: _rb1, rb2, .. rbn_.
Now we compute their intersection going through present keys one by one.
If for a key _k_ there are _n_ containers – we compute the intersection as usual:
```
c := c1.and(c2)
c = c.iand(c3)
// ... looping up until cn
```

However if for a key _k'_ we found less than _n_ containers (by calling `PopNextContainers()`) we know that for one bitmap there was no container at _k'_ present. As we know intersection with an empty set always results in an empty set. Therefore we can skip calling intersection code and skip adding any container at this key to the resulting bitmap – it would be empty after all.

parallel.go
lemire: If what you say is correct, then why do you need

´´´if len(ck.containers) == bitmapCount {´´´

lemire:
I am sure that the intersection code is correct but I think you can have n input bitmaps, aggregate k (less than n) containers and then discard the computed containers. Else why do you need to check the count at the end of your function?

parallel.go
maciej: `h.PopNextContainers()` outputs all containers it finds for the next key for all input bitmaps. It has no logic. Checking if it's _n_ or less than _n_ belongs to the caller (here `ParAnd`).

maciej:
I don't know where the misunderstanding comes from.
`if len(ck.containers) == bitmapCount` happens at the beginning of the intersection loop, not at the end.
Even though `andFunc` is defined and started earlier it does nothing until a `multipleContainers` message is passed on the channel `inputChan <- ck` (around line 293), 2 lines below the `len` check.

lemire:
Ok. What is the computational complexity of the intersection in terms of the number of containers? It should be O(N) where N is the total number of containers. I think that your computational complexity is more like O(N log(B)) where B is the number of bitmaps.

Maybe I am not getting through my message...

But anyhow, I am convinced that this is correct and good code, so we shall merge.

My point is that I think you are leaving quite a bit of performance on the table in some cases where there are many containers.

It would be great to have some version of your benchmarks in the future.

lemire:
Make sure you sync because golint revealed some minor issues.

lemire:
I have opened a separate issue to address my concerns regarding the optimality of the implementation.

https://github.com/RoaringBitmap/roaring/pull/109

lemire:
Let me elaborate some more with an example.

Suppose you have an empty bitmap and 10,000 other bitmaps. What is your code going to to do? For the intersection, it will do a lot of work for nothing.

lemire:
The issue where I describe a possible improvement is https://github.com/RoaringBitmap/roaring/issues/110

lemire:
Just so we are clear: this is not a *fault* in your implementation that is concerning me. Rather, it is just that I think we can do better *in some cases* with code that could be just as clean.

PR #111, opened 2017-10-22
Fixes an issue with ParAnd producing bitmaps with empty containers

maciej:
This commit fixes an issue with `ParAnd` producing bitmaps that
contain empty containers, which AFAIK are invalid.

@lemire sorry for bothering you with a fixup to the recent PR. It's a case I haven't thought of before we had a discussion on intersections.


lemire:
Great.

PR #112, opened 2017-10-24
travis: update go versions

dvrkps:


cornelk:
just as an extra info: x will use the latest minor version

lemire:
👍

PR #119, opened 2017-10-28
FastAggregation: ensure COW is respected

maciej:
This PR fixes two issues with `FastOr` method. Both of them are related to `func (*Bitmap) lazyOR(*Bitmap) *Bitmap)`:
* in the case when containers where appended to the resulting bitmap copy-on-write marker was not carried over, which in some cases resulted in writing to a container that requires copy on write
* if a bitmap container was encountered in the receiver `Bitmap` copy-on-write marker was not respected


lemire:
Looks good.

PR #120, opened 2017-10-28
Parallel: remove redundant container copying

maciej:
`bitmapContainerHeap` always returned writeable containers even though there was is need for that. This PR removes the unnecessary cloning of containers.


lemire:
Good.

PR #121, opened 2017-10-28
Bitmap.FromBuffer

### LINKED ISSUE #116: No-copy bitmap creation on existing memory area
It should be possible to create a `roaring.Bitmap` from a memory buffer (a `byte[]`) containing a memory-mapped serialised Roaring Bitmap, if the [RoaringFormatSpec](https://github.com/RoaringBitmap/RoaringFormatSpec) was used. Unlike `Bitmap.ReadFrom(io.Reader)` it would not copy the input data.

I think the method might not be portable across every architecture / host system. It would use [reflect.SlideHeader](https://golang.org/pkg/reflect/#SliceHeader) to construct roaring data structures (mostly container data).

maciej:
This PR introduces a method that allows to create a Bitmap from a byte buffer (`[]byte`) containing a serialized Roaring Bitmap with minimal memory copying.

Closes #116.

### DIFF
--- a/roaring.go
+++ b/roaring.go
@@ -71,6 +71,20 @@ func (rb *Bitmap) ReadFrom(stream io.Reader) (int64, error) {
 	return rb.highlowcontainer.readFrom(stream)
 }
 
+// FromBuffer creates a bitmap from its serialized version stored in buffer
+//
+// The format specification is available here:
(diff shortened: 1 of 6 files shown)


lemire:
This looks pretty good to me. I will maybe tweak a tiny bit the wording after merging.

PR #124, opened 2017-11-07
bitmapContainer.toArrayContainer: remove double slice allocation

maciej:
Calling `newArrayContainerCapacity(bc.cardinality)` in unnecessary as `ac.loadData(bc)` re-creates the `ac.content`.


lemire:
Good catch!!!

PR #128, opened 2017-12-29
Generic bitmapContainer.readFrom() function should compute cardinality

maciej:
Fix `bitmapContainer.readFrom()` to compute cardinality just like
its _littleendian_ counterpart.


maciej:
TravisCI for some reason was looking for the branch in _RoaringBitmap/roaring_ repository instead of _maciej/roaring_. To fix it I pushed a branch with the same name to the upstream repo. Now the Travis checks pass.

lemire:
Looks like a no-brainer regarding merging.

lemire:
Done.

Issue #122, opened 2017-10-28
Align internal run container representation with serialized format


maciej:
Currently run container data is represented as an `[]interval16`, which is defined as
```go
type interval16 struct {
	start uint16
	last  uint16
}
```
This differs from the representation of a serialized bitmap as described [here](https://github.com/RoaringBitmap/RoaringFormatSpec). In effect it makes it hard/impossible to implement no-copy bitmap reading. I propose changing this representation to address that.

See also #116 and discussion in #37.

lemire:
👍

maciej:
I've started working on this in [maciej/rle-branch](https://github.com/maciej/roaring/tree/rle-len). My implementation plan is as follows:
- [x] introduce _length_ field to the existing `interval16` struct
- [ ] rewrite `interval16` struct creation in test code to be using `newInterval16FromStartLast` function instead
- [ ] progressively fix tests and code up to a point where tests pass with the bigger `interval16`
- [ ] remove `last` field from `interval16`
- [ ] fix code and tests that were broken by that removal
- [ ] fix serialization

(I'll keep updating this list as I progress with work)

lemire:
Excellent.

glycerine:
Well I suppose it would be the acme of foolishness to point out that it would be faster to just add another serialization format option that does match the [start, last] format and do reader-makes-right lazy correction only if needed (here I duck the flying tomatoes :)

Either way I wish this effort well. I know that it is a big effort.  I found it much easier to reason about [start, last] when getting correctness, as the corner cases are many and subtle. Revisit my comments on crossing bin boundaries in #37 and be sure to add tests cases for that situation if it does turn out out to be a hazard.

When you are testing for correctness, be sure to run extra trials by extending the `n` and `ntrial` values in the test suite (reference [1] and [2] below).  Running these with big values caught alot of corner cases in my origin implementation, but are too slow to leave on for every day/every commit test runs.

[1] https://github.com/RoaringBitmap/roaring/blob/master/rle16_test.go#L325

[2] https://github.com/RoaringBitmap/roaring/blob/master/rlei_test.go#L1616

maciej:
@glycerine thank you for the pieces of advice! As I'll move forward with the code I'll certainly use them. :-)

[...]

PR #129, opened 2018-01-03
FromBuffer: allocate roaring array slices upfront

maciej:
When using the updated RLE format (see #122) the main cost of constructing a large
bitmap with FromBuffer becomes growing the roaring array internal slices when
appending containers.

This commits fixes that. Now the `keys`, `containers` and `needCopyOnWrite` slices are
preallocated.


lemire:
This seems safe and good, and it does not seem to have any dependencies on anything else... so I will merge this without further review.

maciej:
:+1:

PR #131, opened 2018-01-16
fromBuffer() changes

maciej:
This PR introduces two changes to `roaringArray.fromBuffer` function:

* when creating the `isRun` bitmap instead looping over `unit64`s (and the remainder of bytes)  I "cast" the `isRun` bitmap to a `[]byte` slice using the `uint64SliceAsByteSlice` function and use the built-in `copy` fun to copy into the bitmap. I can observe a slight performance improvement on production systems due to that change

* a less significant change removes usage of `appendContainer` method and instead populates the `roaringArray` slices directly. I don't know if this has any impact on performance. Taking it on face value we eliminate the slice length change and the assignment. I have no measured if it has any impact on performance TBH, it'd likely be very insignificant if anything. Frankly, I simply prefer that way of expressing the operation in code. :-)

### DIFF
--- a/roaringarray.go
+++ b/roaringarray.go
@@ -571,13 +571,16 @@ func (ra *roaringArray) fromBuffer(buf []byte) (int64, error) {
 	}
 
 	// Allocate slices upfront as number of containers is known
-	ra.containers = make([]container, 0, size)
-	ra.keys = make([]uint16, 0, size)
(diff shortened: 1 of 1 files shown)


maciej:
I'll close this PR and reopen another one. Hopefully this will fix travis.

PR #132, opened 2018-01-16
roaringArray.fromBuffer() changes

maciej:
This PR introduces two changes to `roaringArray.fromBuffer` function:

* when creating the `isRun` bitmap instead looping over `unit64`s (and the remainder of bytes)  I "cast" the `isRun` bitmap to a `[]byte` slice using the `uint64SliceAsByteSlice` function and use the built-in `copy` fun to copy into the bitmap. I can observe a slight performance improvement on production systems due to that change

* a less significant change removes usage of `appendContainer` method and instead populates the `roaringArray` slices directly. I don't know if this has any impact on performance. Taking it on face value we eliminate the slice length change and the assignment. I have no measured if it has any impact on performance TBH, it'd likely be very insignificant if anything. Frankly, I simply prefer that way of expressing the operation in code. :-)

### DIFF
--- a/roaringarray.go
+++ b/roaringarray.go
@@ -541,7 +541,7 @@ func (ra *roaringArray) fromBuffer(buf []byte) (int64, error) {
 	pos += 4
 	var size uint32 // number of containers
 	haveRunContainers := false
-	var isRun *bitmapContainer
+	var isRunBitmap []byte
 
 	// cookie header
 	if cookie&0x0000FFFF == serialCookie {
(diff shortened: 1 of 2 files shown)


lemire:
If you want to optimize... then you don't need isRun at all!!! No copy is necessary, at all, which is surely going to be faster (and avoid generating garbage)... Have a look at the equivalent C code...

https://github.com/RoaringBitmap/CRoaring/blob/8f036c1e7ebfa0c20bee56cdbf241f7f9413d7b6/src/roaring_array.c#L578-L605

The bitset is already there, in bytes, we just need to access its bits. So this line (which is already there) is enough...

```
by := buf[pos : pos+bytesToRead]
```

Then we just need to do something like this to access the bit values...

```
if (by[k / 8] & (1 << (k % 8))) != 0 {
}
```

(Appropriate casting is probably needed to get this to compile cleanly.)

Similarly, the serialization code can probably be optimized.

lemire:
I also agree with the append stuff... direct access is cleaner.

lemire:
So I suggest you have a look at my proposal for a deeper optimization resulting in fewer lines and faster code.

maciej:
@lemire You're right! I'm reimplement it without using the `bitmapContainer` at all.

maciej:
@lemire give me a few more days to complete it. I'll rewrite the container type (_run_ or not) detection to be copy-less.

maciej:
@lemire done! now the `isRun` bitmap is not copied.

lemire:
Great stuff. Let us merge.

Issue #113, opened 2017-10-25
Benchmark Or functions computing union of ~100 large sparse bitmaps


maciej:
The goal of this task is to write benchmark that will take ~100 generated sparse bitmaps and compute their unions using existing aggregation methods: `ParOr`, `FastOr`, …

lemire:
Right. I have written a naive benchmark... https://github.com/lemire/gobitmapbenchmark It is not good enough to assess performance seriously.

One needs to balance such benchmarks with benchmarks where we have long runs of values.  That is, we need to test all container types, and their interactions.

We have a fairly complete benchmark that we use for both the C and Java version : 

https://github.com/RoaringBitmap/CBitmapCompetition

You can find the data (as text files) there...

https://github.com/RoaringBitmap/CRoaring/tree/edc6fe04a26c43267b9e87efeb1f18d7945f8ae4/benchmarks/realdata

It tests many edge cases. 

It is not hard to implement.

Note that we can benchmark against the C implementation : 
https://github.com/RoaringBitmap/gocroaring

maciej:
I think #137 covers that well enough.

lemire:
Indeed.

PR #133, opened 2018-01-16
Parallel operation changes

maciej:
This PR introduces two changes to parallel operations:
* it removes the unnecessary container pointer from `bitmapContainerKey` struct. This has a surprisingly significant impact on performance in the private benchmark[1] I'm using.
* it pools `[]container` slice passed to `orFunc` in `ParOr`. The measured perf. impact is not as significant as in the first change, but it's also positive

[1] I hope, hope to make some progress on #113 soon! :-)

### DIFF
--- a/parallel.go
+++ b/parallel.go
@@ -9,10 +9,9 @@ import (
 var defaultWorkerCount = runtime.NumCPU()
 
 type bitmapContainerKey struct {
-	bitmap    *Bitmap
-	container container
-	key       uint16
-	idx       int
+	key    uint16
+	idx    int
+	bitmap *Bitmap
 }
 
 type multipleContainers struct {
@@ -43,7 +42,7 @@ func (h *bitmapContainerHeap) Pop() interface{} {
 	old := *h
 	n := len(old)
 	x := old[n-1]
-	*h = old[0: n-1]
+	*h = old[0 : n-1]
 	return x
 }
 
@@ -54,15 +53,14 @@ func (h bitmapContainerHeap) Peek() bitmapContainerKey {
 func (h *bitmapContainerHeap) popIncrementing() (key uint16, container container) {
 	k := h.Peek()
 	key = k.key
-	container = k.container
+	container = k.bitmap.highlowcontainer.containers[k.idx]
 
 	newIdx := k.idx + 1
 	if newIdx < k.bitmap.highlowcontainer.size() {
 		k = bitmapContainerKey{
-			k.bitmap,
-			k.bitmap.highlowcontainer.containers[newIdx],
 			k.bitmap.highlowcontainer.keys[newIdx],
 			newIdx,
+			k.bitmap,
 		}
 		(*h)[0] = k
 		heap.Fix(h, 0)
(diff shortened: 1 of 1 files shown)


lemire:
Merging.

If you get around to it, I'd like to know why (you think) the container point hurts the performance so much. Is that a caching issue?

lemire:
+1 on the progress regarding the benchmarking!

maciej:
@lemire I think the answer might have little to do with CPU caches.

I've used Apple's Xcode Instrumentation tool to get traces for both old and new code. It runs a 100 iterations of a set expression on digital advertising data on top of roaring (the private benchmark I'm using).

new:
```
6249	6249.0ms    8.2%	181.0	 	       github.com/RoaringBitmap/roaring.ParOr
5943	5943.0ms    7.8%	1473.0	 	        github.com/RoaringBitmap/roaring.(*bitmapContainerHeap).Next
4450	4450.0ms    5.8%	146.0	 	         github.com/RoaringBitmap/roaring.(*bitmapContainerHeap).popIncrementing
4300	4300.0ms    5.6%	2129.0	 	          container/heap.Fix
2171	2171.0ms    2.8%	2021.0	 	           container/heap.down
150	150.0ms    0.1%	4.0	 	            github.com/RoaringBitmap/roaring.(*bitmapContainerHeap).Swap
146	146.0ms    0.1%	41.0	 	             runtime.writebarrierptr
```

old:
```
12242	12242.0ms    9.2%	450.0	 	       github.com/RoaringBitmap/roaring.ParOr
11729	11729.0ms    8.8%	2476.0	 	        github.com/RoaringBitmap/roaring.(*bitmapContainerHeap).PopNextContainers
8526	8526.0ms    6.4%	657.0	 	         github.com/RoaringBitmap/roaring.(*bitmapContainerHeap).PopIncrementing
7782	7782.0ms    5.8%	3239.0	 	          container/heap.Fix
4543	4543.0ms    3.4%	2658.0	 	           container/heap.down
1885	1885.0ms    1.4%	1127.0	 	            github.com/RoaringBitmap/roaring.(*bitmapContainerHeap).Swap
758	758.0ms    0.5%	166.0	 	             runtime.typedmemmove
592	592.0ms    0.4%	295.0	 	              runtime.bulkBarrierPreWrite
79	79.0ms    0.0%	13.0	 	          runtime.typedmemmove
66	66.0ms    0.0%	36.0	 	           runtime.bulkBarrierPreWrite
```

Now, comparing the traces Swap functions on `bitmapContainerHeap` it looks like it takes Go 10x long to move the structure containing `container` compared to the one that does not contain it.

` runtime.typedmemmove` seems to be taking a bulk of that time. That's all what I can tell looking at the traces. What `runtime.typedmemmove` does and why is it taking so much CPU time – I don't know.

maciej:
There is surprisingly a lot happening in a one line function when you see it's [x86 assembly code](https://gist.github.com/maciej/089087384b8b4cc275dea86c9a58ffde).

lemire:
I would expect the containers to be pointers to structs, wouldn't you? If so, then you had two pointers, effectively, an int and a 16-bit int. So... what? 4x8 = 32 bytes? Ok. So now you streamlined to int, 16-bit and one pointer. So... what? 3x8 = 24 bytes... it is not the difference of 8 bytes (or whatever) that would make much of a difference.

Maybe there is an impact on garbage collection?

maciej:
From what I understand the Go GC design (and I hardly do) the write barriers are necessary for the GC to function: https://github.com/golang/proposal/blob/master/design/17503-eliminate-rescan.md

lemire:
I'm annoyed that such a small difference can make a 10x perf difference and I don't have a clue as to why that is.

maciej:
https://groups.google.com/forum/#!topic/golang-nuts/0ZlHR30p2EQ let's hope someone on _golang-nuts_ group will be kind enough to explain that

PR #139, opened 2018-02-02
Generic aggregations tests

maciej:
This PR introduces generic aggregation tests and removes the duplicate test code that was introduced with parallel aggregations.

I also rewrote the test code to use plain Go >1.7 sub-tests. I find them to be cleaner and more "idiomatic Go" compared to Go Convey.


lemire:
Looks good to me. Our tests are written using all sorts of idioms, it

PR #143, opened 2018-02-05
Eliminate unnecessary Bitmap allocation in (*Bitmap).lazyOR

maciej:
Fixes #138

Benchmark on the "real" dataset show a significant improvement:

```
benchmark                                            old ns/op     new ns/op     delta
BenchmarkRealDataFastOr/census-income_srt-4          841099        736937        -12.38%
BenchmarkRealDataFastOr/census-income-4              1248805       1157445       -7.32%
BenchmarkRealDataFastOr/census1881_srt-4             1992651       1295663       -34.98%
BenchmarkRealDataFastOr/census1881-4                 2797780       2096418       -25.07%
BenchmarkRealDataFastOr/dimension_003-4              50486750      2174251       -95.69%
BenchmarkRealDataFastOr/dimension_008-4              17056898      904640        -94.70%
BenchmarkRealDataFastOr/dimension_033-4              1246254       794440        -36.25%
BenchmarkRealDataFastOr/uscensus2000-4               5387897       2702060       -49.85%
BenchmarkRealDataFastOr/weather_sept_85_srt-4        713331        505671        -29.11%
BenchmarkRealDataFastOr/weather_sept_85-4            5248820       4952945       -5.64%
BenchmarkRealDataFastOr/wikileaks-noquotes_srt-4     567737        276666        -51.27%
BenchmarkRealDataFastOr/wikileaks-noquotes-4         906403        613208        -32.35%
```

PR #145, opened 2018-02-06
Manually inline setBitmapRange to *bitmapContainer.lazyIOR function

maciej:
Even though it adds code duplication the performance benchmarks make
this approach interesting:

```
benchmark                                            old ns/op     new ns/op     delta
BenchmarkRealDataFastOr/census-income_srt-8          617026        395117        -35.96%
BenchmarkRealDataFastOr/census1881_srt-8             1051061       935098        -11.03%
BenchmarkRealDataFastOr/weather_sept_85_srt-8        404418        297734        -26.38%
BenchmarkRealDataFastOr/wikileaks-noquotes_srt-8     214869        190827        -11.19%
```

Note that the benchmarks were done against FastOr in #143

### DIFF
--- a/bitmapcontainer.go
+++ b/bitmapcontainer.go
@@ -347,13 +347,28 @@ func (bc *bitmapContainer) lazyIOR(a container) container {
 		if x.isFull() {
 			return x.clone()
 		}
-		// TODO : implement efficient in-place lazy OR to bitmap
-		for i := range x.iv {
-			setBitmapRange(bc.bitmap, int(x.iv[i].start), int(x.iv[i].last())+1)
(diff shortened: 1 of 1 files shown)


lemire:
I think that Go allows for manual inlining... so this looks good to me.

maciej:
Although the Go compiler does function inlining it does not inline functions that are "complex" (that for example contain loops[1]) like `setBitmapRange`. Since in practice the time needed to invoke `setBitmapRange` function was roughly similar to the time `setBitmapRange` executed it made sense to copy the code into the `lazyIOR` function to improve performance (at a cost of code duplication).

[1] https://github.com/golang/go/wiki/CompilerOptimizations#function-inlining

lemire:
+1

PR #156, opened 2018-03-12
Bitmap.Clear() retains preallocated slices for faster future ops

steveyen:
Before this commit, Clear() would throw away any allocated slices, and
this change has Clear() instead resize slice lengths to 0, keeping the
underlying slice capacity.  Future operations that append() to the
slices might then see performance wins by avoiding some
memory allocations.

fromBuffer() is also enhanced to potentially reuse preallocated slice capacity.

An application that reuses a Bitmap instance via Clear() may see more performance gains than before.

### DIFF
--- a/roaring.go
+++ b/roaring.go
@@ -146,9 +146,10 @@ func New() *Bitmap {
 	return &Bitmap{}
 }
 
-// Clear removes all content from the Bitmap and frees the memory
+// Clear resets the Bitmap to be logically empty, but may retain
+// some memory allocations that may speed up future operations
 func (rb *Bitmap) Clear() {
-	rb.highlowcontainer = *newRoaringArray()
+	rb.highlowcontainer.clear()
 }
 
 // ToArray creates a new slice containing all of the integers stored in the Bitmap in sorted order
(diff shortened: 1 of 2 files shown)


lemire:
c.c. @maciej

lemire:
My main concern is that the performance gains are not measured.

steveyen:
> My main concern is that the performance gains are not measured.

Hi @lemire -- indeed, I'll get around to adding a benchmark as part of this.  I do see nice gains in application-level benchmarks (using roaring as part of a new indexing engine in the bleve full-text indexing library).

cheers,
steve

lemire:
@steveyen 

Thanks. To be clear, I have no objection to the code change per se. It is fine.

lemire:
I am going to merge this, but it would be really nice if you considered adding a corresponding benchmark.

lemire:
See https://github.com/RoaringBitmap/roaring/issues/158

steveyen:
doing some cross-linking -- https://github.com/RoaringBitmap/roaring/pull/159

PR #164, opened 2018-04-03
Faster ParOr

### LINKED ISSUE #140: Parallel aggregation: split work by ranges of container keys
Currently all work handed of to goroutines handling container-level aggregation is prepared in the caller goroutine  by picking a new _work unit_ from a heap. Private benchmarks (I hope to replicate the results using the public datasets) show that operating on the heap takes a fair amount of time limiting the parallelism level and taking CPU time that could otherwise be spent on the aggregation. Especially `ParAnd` could benefit from a solution that would not use the heap, as pointed out in #110.

Here is an informal description of how a `ParOr` could look like: for the given list of bitmaps _B_ lets take the smallest _Ks_ and largest _Ke_ key. Let's split the _[Ks, Ke]_ range into _N_ non-overlapping ranges. _Ks_ and _Ke_ can be computed in O(B) time.  _N_ can be determined based on the number of available CPU cores/goroutines (x * CPU count, I'd start with x anywhere between 2 and 16) and size of the range (perhaps it does not make sense to make the range 1 or 2 container long). This will be subject to fine-tuning.
Having those subranges we can then operate on them in the same manner as `FastOR` function would, but limited on a range of keys.
Results of the work units could be easily concatenated into a single `roaringArray`.

It's an open question if that would be faster, however I think it's a promising approach. :-) The idea is somehow inspired by the discussions on [RoaringBitmap#42](https://github.com/RoaringBitmap/RoaringBitmap/issues/42) and my limited understanding of how splitterator works.

maciej:
This PR introduces a faster _ParOr_. The algorithm is described informally in #140.

Benchmark data shows significant improvement compared to previous algorithm.

Benchmarks run on 15' 2017 MacBook Pro Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
```
benchmark                                           old ns/op     new ns/op     delta
BenchmarkRealDataParOr/census-income_srt-8          270110        198445        -26.53%
BenchmarkRealDataParOr/census-income-8              383111        302936        -20.93%
BenchmarkRealDataParOr/census1881_srt-8             679945        404061        -40.57%
BenchmarkRealDataParOr/census1881-8                 701318        606377        -13.54%
BenchmarkRealDataParOr/dimension_003-8              7206730       2208563       -69.35%
BenchmarkRealDataParOr/dimension_008-8              3016907       959910        -68.18%
BenchmarkRealDataParOr/dimension_033-8              469810        201631        -57.08%
BenchmarkRealDataParOr/uscensus2000-8               1882977       1081049       -42.59%
BenchmarkRealDataParOr/weather_sept_85_srt-8        382546        158591        -58.54%
BenchmarkRealDataParOr/weather_sept_85-8            1255767       1240276       -1.23%
BenchmarkRealDataParOr/wikileaks-noquotes_srt-8     286285        123959        -56.70%
BenchmarkRealDataParOr/wikileaks-noquotes-8         328856        166740        -49.30%
```

The previous algorithm is exposed in the public API as ParHeapOr somehow resembling the convention in `fastaggregation.go`.

Resolves #140

### DIFF
--- a/aggregation_test.go
+++ b/aggregation_test.go
@@ -215,7 +215,7 @@ func testAggregations(t *testing.T,
 		bigxor := Xor(Xor(rb1, rb2), rb3)
 
 		if or != nil && !or(rb1, rb2, rb3).Equals(rb1) {
-			t.Error()
+			t.Errorf("Expected bitmap of cardinality %d, got %d", rb1.GetCardinality(), or(rb1, rb2, rb3).GetCardinality())
 		}
 
 		if and != nil && !and(rb1, rb2, rb3).Equals(bigand) {
@@ -257,7 +257,7 @@ func testAggregations(t *testing.T,
 		bigxor := Xor(Xor(rb1, rb2), rb3)
 
 		if or != nil && !or(rb1, rb2, rb3).Equals(rb1) {
-			t.Error()
+			t.Errorf("Expected bitmap of cardinality %d, got %d", rb1.GetCardinality(), or(rb1, rb2, rb3).GetCardinality())
 		}
 
 		if and != nil && !and(rb1, rb2, rb3).Equals(bigand) {
@@ -279,6 +279,14 @@ func TestParAggregations(t *testing.T) {
 	}
 
 	testAggregations(t, andFunc, orFunc, nil)
+}
+
+func TestParHeapAggregations(t *testing.T) {
+	orFunc := func(bitmaps ...*Bitmap) *Bitmap {
+		return ParHeapOr(0, bitmaps...)
+	}
+
+	testAggregations(t, nil, orFunc, nil)
 }
 
 func TestFastAggregations(t *testing.T) {
--- a/arraycontainer.go
+++ b/arraycontainer.go
@@ -803,19 +803,12 @@ func (ac *arrayContainer) negateRange(buffer []uint16, startIndex, lastIndex, st
 	}
 }
 
-func min(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 func (ac *arrayContainer) isFull() bool {
 	return false
 }
 
 func (ac *arrayContainer) andArray(value2 *arrayContainer) container {
-	desiredcapacity := min(ac.getCardinality(), value2.getCardinality())
+	desiredcapacity := minOfInt(ac.getCardinality(), value2.getCardinality())
 	answer := newArrayContainerCapacity(desiredcapacity)
 	length := intersection2by2(
 		ac.content,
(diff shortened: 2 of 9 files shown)


maciej:
Whoops! I haven't done proper testing before pushing this! Sorry about that! It should be a trivial edge case to fix, let me do just that!

maciej:
Bizarrely those tests fail on Travis and pass on my local machine.
```
$ go test
--- FAIL: TestParAggregations (0.05s)
    --- FAIL: TestParAggregations/advanced_case (0.04s)
    	aggregation_test.go:218: Expected bitmap of cardinality 857144, got 758346
    --- FAIL: TestParAggregations/advanced_case_with_runs (0.02s)
    	aggregation_test.go:260: Expected bitmap of cardinality 635287, got 536489
```

I've tested it against Go 1.10.1, 1.9.5 and 1.9.4. In each case the cardinality is the same.

maciej:
OK, I'm able to reproduce it on `linux/arm` (Raspberry Pi)
```
◦  go test -run TestParAgg                                                                                         220:47:19
--- FAIL: TestParAggregations (0.37s)
    --- FAIL: TestParAggregations/advanced_case (0.20s)
    	aggregation_test.go:218: Expected bitmap of cardinality 857144, got 477478
    --- FAIL: TestParAggregations/advanced_case_with_runs (0.16s)
    	aggregation_test.go:260: Expected bitmap of cardinality 635287, got 255621
FAIL
exit status 1
FAIL	github.com/RoaringBitmap/roaring	0.383s
```

lemire:
If you’d like, I can do a code review now or do you prefer I wait?

lemire:
cc @richardstartin

maciej:
@lemire please wait. I messed up chunk size calculation. Due to some tuning constants the tests passed on a machine with `NumCPU = 8` (4 core Mac with hyper threading), but failed on eg. `parallelism = 4`. I'll improve the test coverage to check against that and get back to you!

Unfortunately that only means some of the benchmark results were wrong. I won't post the updated results just yet, I want to be 100% sure they are correct.

richardstartin:
One thing to pay attention to might be `uscenus2000` - that's over 1ms, and is under 100us in a [Java PR](https://github.com/RoaringBitmap/RoaringBitmap/pull/234), on my slightly lower spec machine. I would be interested to see these two approaches implemented in the same language, in each language, to better understand the tradeoff between up front work and balance.

maciej:
@richardstartin thank you for sharing the dataset characteristics! These are very valuable!
Yes, I think it'd make sense to have all _ParOr_ implementations compared against one another at some point!
For now I would not rely on the numbers I posted above. They are wrong.

maciej:
I'm closing this PR. I'll do #168 and #169 and issue another PR once I'm confident that it's correct.

PR #170, opened 2018-04-04
Fuzzing framework for serialization

dgryski:


lemire:
Fantastic.

maciej:
@dgryski correct me if I'm wrong, but `crashproneinput*.bin` files are not used as part of this test, right?

PR #177, opened 2018-04-10
TestParAggregations with different parallelism levels

### LINKED ISSUE #168: TestParAggregations should test different parallelism levels
Currently `TestParAggregations` runs tests in `ParAnd` & `ParOr` only with parallelism level set to 0. I'd suggest testing at least 3 levels instead: 1, 2, 4. That would avoid subtle bugs that manifest themselves only on one of those levels as seen here #164.

maciej:
Fixes #168

### DIFF
--- a/aggregation_test.go
+++ b/aggregation_test.go
@@ -4,6 +4,7 @@ package roaring
 
 import (
 	"testing"
+	"fmt"
 )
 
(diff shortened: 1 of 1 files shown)


lemire:
Only tests, so it should be safe to merge.

Issue #169, opened 2018-04-04
Validate bitmap cardinality in BenchmarkRealData* functions


maciej:
I think it makes sense to validate bitmap cardinality in `BenchmarkRealData*` functions. In some rare cases the `aggregation_test.go` test suite does not reveal subtle bugs that could be introduced while working on wide union functions. Typically such work is done to improve performance and benchmarks are run as part of it. So even though `BenchmarkRealData*` is not part of test suite I think I'd allow us to catch some bugs.
I'd also help me avoiding publishing benchmarks that are plain wrong as in #164. :)

lemire:
Agreed. Marking as "help wanted".

richardstartin:
Might you find some of these with invariance testing?

maciej:
@richardstartin very good point!

PR #178, opened 2018-04-10
Faster ParOr

### LINKED ISSUE #140: Parallel aggregation: split work by ranges of container keys

maciej:
This commit introduces a faster _ParOr_. The algorithm is described informally in #140.

Benchmark data shows significant improvement compared to previous algorithm.

Benchmarks run on 15' 2017 MacBook Pro Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
```
benchmark                                           old ns/op     new ns/op     delta
BenchmarkRealDataParOr/census-income_srt-8          269673        194486        -27.88%
BenchmarkRealDataParOr/census-income-8              395797        310505        -21.55%
BenchmarkRealDataParOr/census1881_srt-8             723821        402359        -44.41%
BenchmarkRealDataParOr/census1881-8                 718247        573550        -20.15%
BenchmarkRealDataParOr/dimension_003-8              7466247       2197374       -70.57%
BenchmarkRealDataParOr/dimension_008-8              3062717       833368        -72.79%
BenchmarkRealDataParOr/dimension_033-8              474489        302080        -36.34%
BenchmarkRealDataParOr/uscensus2000-8               1912128       1092854       -42.85%
BenchmarkRealDataParOr/weather_sept_85_srt-8        370585        178577        -51.81%
BenchmarkRealDataParOr/weather_sept_85-8            1272372       1142709       -10.19%
BenchmarkRealDataParOr/wikileaks-noquotes_srt-8     290843        123167        -57.65%
BenchmarkRealDataParOr/wikileaks-noquotes-8         330745        152330        -53.94%
```

The previous algorithm is exposed in the public API as `ParHeapOr` somehow resembling the convention in `fastaggregation.go`.

Resolves #140

It's the second revision of the pull request. It fixed bugs found in #164

I've compared the cardinalities on /Real data/ dataset https://gist.github.com/maciej/add052d10c5edad2eb74eceda7484f07 semin-manually. They are the same, so I'm leaving #169 for later

### DIFF
--- a/parallel.go
+++ b/parallel.go
@@ -2,9 +2,9 @@ package roaring
 
 import (
 	"container/heap"
+	"fmt"
 	"runtime"
 	"sync"
-	"fmt"
 )
 
 var defaultWorkerCount = runtime.NumCPU()
@@ -386,7 +386,7 @@ func ParOr(parallelism int, bitmaps ...*Bitmap) *Bitmap {
 
 	if chunkCount*chunkSize < int(keyRange) {
 		// it's fine to panic to indicate an implementation error
-		panic(fmt.Sprintf("invariant check failed: chunkCount * chunkSize >= keyRange, %d * %d < %d", chunkCount, chunkSize, keyRange))
+		panic(fmt.Sprintf("invariant check failed: chunkCount * chunkSize < keyRange, %d * %d < %d", chunkCount, chunkSize, keyRange))
 	}
 
 	chunks := make([]*roaringArray, chunkCount)
@@ -417,7 +417,7 @@ func ParOr(parallelism int, bitmaps ...*Bitmap) *Bitmap {
 		for i := 0; i < chunkCount; i++ {
 			spec := parChunkSpec{
 				start: uint16(int(lKey) + i*chunkSize),
-				end:   uint16(minOfInt((i+1)*chunkSize-1+int(lKey), int(hKey))),
+				end:   uint16(minOfInt(int(lKey)+(i+1)*chunkSize-1, int(hKey))),
 				idx:   int(i),
 			}
 			chunkSpecChan <- spec
@@ -504,14 +504,14 @@ func lazyOrOnRange(ra1, ra2 *roaringArray, start, last uint16) *roaringArray {
 			if key1 < key2 {
 				answer.appendCopy(*ra1, idx1)
 				idx1++
-				if idx1 >= length1 {
+				if idx1 == length1 {
 					break
 				}
 				key1 = ra1.getKeyAtIndex(idx1)
 			} else if key1 > key2 {
 				answer.appendCopy(*ra2, idx2)
 				idx2++
-				if idx2 >= length2 {
+				if idx2 == length2 {
 					break
 				}
 				key2 = ra2.getKeyAtIndex(idx2)
@@ -521,7 +521,7 @@ func lazyOrOnRange(ra1, ra2 *roaringArray, start, last uint16) *roaringArray {
 				answer.appendContainer(key1, c1.lazyOR(ra2.getContainerAtIndex(idx2)), false)
 				idx1++
 				idx2++
-				if idx1 >= length1 || idx2 >= length2 {
+				if idx1 == length1 || idx2 == length2 {
 					break
 				}
 
@@ -532,21 +532,23 @@ func lazyOrOnRange(ra1, ra2 *roaringArray, start, last uint16) *roaringArray {
 	}
 
 	if idx2 < length2 {
-		key2 := ra2.getKeyAtIndex(idx2)
+		key2 = ra2.getKeyAtIndex(idx2)
 		for key2 <= last {
 			answer.appendContainer(key2, ra2.getContainerAtIndex(idx2), true)
 			idx2++
-			if idx2 >= length2 {
+			if idx2 == length2 {
 				break
 			}
 			key2 = ra2.getKeyAtIndex(idx2)
 		}
-	} else if idx1 < length1 {
-		key1 := ra1.getKeyAtIndex(idx1)
+	}
+
+	if idx1 < length1 {
+		key1 = ra1.getKeyAtIndex(idx1)
 		for key1 <= last {
 			answer.appendContainer(key1, ra1.getContainerAtIndex(idx1), true)
 			idx1++
-			if idx1 >= length1 {
+			if idx1 == length1 {
 				break
 			}
 			key1 = ra1.getKeyAtIndex(idx1)
@@ -603,7 +605,7 @@ func lazyIOrOnRange(ra1, ra2 *roaringArray, start, last uint16) *roaringArray {
 	}
 
 	if idx2 < length2 {
-		key2 := ra2.getKeyAtIndex(idx2)
+		key2 = ra2.getKeyAtIndex(idx2)
 		for key2 <= last {
 			ra1.appendContainer(key2, ra2.getContainerAtIndex(idx2), true)
 			idx2++
--- a/parallel.go
+++ b/parallel.go
@@ -376,13 +376,8 @@ func ParOr(parallelism int, bitmaps ...*Bitmap) *Bitmap {
 		chunkCount = int(keyRange)
 	} else {
 		chunkCount = parallelism * 4
-		if int(keyRange)%chunkCount > 0 {
-			chunkSize = (int(keyRange) / chunkCount) + 1
-		} else {
-			chunkSize = int(keyRange) / chunkCount
-		}
+		chunkSize = (int(keyRange) + chunkCount - 1) / chunkCount
 	}
-	//fmt.Printf("cc %d cs %d r %d d %d\n", chunkCount, chunkSize, int(keyRange)%chunkSize, chunkCount*chunkSize-int(keyRange))
 
 	if chunkCount*chunkSize < int(keyRange) {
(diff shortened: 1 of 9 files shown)


maciej:
cc @richardstartin

richardstartin:
Looks great, might be worth looking at uscensus2000 because it's quite an easy problem, the rest are comparable to the Java implementation (allowing for a different language and benchmarking machine).

maciej:
@richardstartin could you link me to the Java ParOr benchmark results? I'd like to create a ticket regardin _uscensus2000_ dataset.

That said in the new `ParOr` there is still space for investigating performance improvements:
* the search for containers within a given key range is linear
* chunk size and count could be fine-tuned even more

richardstartin:
There's a discussion (argument?) about it [here](https://github.com/RoaringBitmap/RoaringBitmap/issues/223), which only applies to the buffer implementation (typically the slower one). There's more discussion [https://github.com/RoaringBitmap/RoaringBitmap/pull/211](here) with various results (so look towards the end, not the beginning). Obviously, benchmarks depend on the machines they run on, but they are surprisingly similar to your results for several cases.

The obvious difference is that the Java implementation makes more effort to balance the tasks. This work takes anything between 10us and 400us. I think there is also a place in the Java project for an implementation assuming uniform key distribution as this one seems to, because it means you can sometimes get away with not doing the preparatory work.

cakey:
Nice! Numbers look great, I'm going to try it on some of our internal suites later!

cakey:
Good and bad news. Your branch seems much much faster (100x?!), (awesome!) bad news is that our tests seem to get different results (off by 1 on some of the cardinalties)... Everything works if I switch back to ParHeapOr on your branch - Trying to work out if our code violates assumptions that don't cause problems in the old function, or if there might be a bug. Working on reproducing either way now.

cakey:
Only seems to happen for parallelism > 1

cakey:
@maciej 

```
	ba1 := []uint32{3585028, 65901253, 143441994, 211160474, 286511937, 356744840, 434332509, 502812785, 576097614, 646557334, 714794241, 775083485, 833704249, 889329147, 941367043}
	ba2 := []uint32{17883, 54494426, 113908938, 174519827, 235465665, 296685741, 357644666, 420192495, 476104304, 523046142, 577855081, 634889665, 692460635, 751350463, 809989192, 863494316, 919127240}

	r1 := roaring.BitmapOf(ba1...)
	r2 := roaring.BitmapOf(ba2...)

	result1 := roaring.ParOr(8, r1, r2)
	result2 := roaring.ParOr(1, r1, r2)
	result3 := roaring.ParHeapOr(1, r1, r2)
	result4 := roaring.ParHeapOr(8, r1, r2)

	fmt.Println(result1.GetCardinality())
	fmt.Println(result2.GetCardinality())
	fmt.Println(result3.GetCardinality())
	fmt.Println(result4.GetCardinality())
```
```
29
30
32
32
```

Possible

lemire:
@maciej  I have not started on a code review, but given that we appear to have a reproducible failing test, maybe you want to have a look at that first?

maciej:
@cakey the issue should now be fixed. Please rerun your tests.

cakey:
@maciej Clean on my end, nice!

maciej:
@richardstartin if I interpret the [RoaringBitmap#211](https://github.com/RoaringBitmap/RoaringBitmap/pull/211) results correctly `parallelOr` on uscensus in Java takes somewhere around 200µs as compared to 1ms here?

richardstartin:
@maciej Yes, but on a different machine, in a different language etc. I wrote an implementation that does uscens2000 in 90us but I dropped the PR. I think the critical difference is balanced task sizes.

parallel.go
```diff
@@ -329,3 +331,289 @@ func ParAnd(parallelism int, bitmaps ...*Bitmap) *Bitmap {
...
+		chunkCount = int(keyRange)
+	} else {
+		chunkCount = parallelism * 4
+		if int(keyRange)%chunkCount > 0 {
+			chunkSize = (int(keyRange) / chunkCount) + 1
+		} else {
+			chunkSize = int(keyRange) / chunkCount
+		}
+	}
+	//fmt.Printf("cc %d cs %d r %d d %d\n", chunkCount, chunkSize, int(keyRange)%chunkSize, chunkCount*chunkSize-int(keyRange))
```
lemire: Do we need this commented out line?

maciej: I left it for debugging issues, but you're right. We can remove it now.

parallel.go
```diff
@@ -329,3 +331,289 @@ func ParAnd(parallelism int, bitmaps ...*Bitmap) *Bitmap {
...
+	if parallelism*4 > int(keyRange) {
+		chunkSize = 1
+		chunkCount = int(keyRange)
+	} else {
+		chunkCount = parallelism * 4
+		if int(keyRange)%chunkCount > 0 {
+			chunkSize = (int(keyRange) / chunkCount) + 1
+		} else {
+			chunkSize = int(keyRange) / chunkCount
+		}
```
lemire: This chunkSize computation seems awfully complicated and branchy. Can't we do ``(keyRange + chunkCount - 1) / chunkCount``?

parallel.go
```diff
@@ -329,3 +331,289 @@ func ParAnd(parallelism int, bitmaps ...*Bitmap) *Bitmap {
...
+
+				key1 = ra1.getKeyAtIndex(idx1)
+				key2 = ra2.getKeyAtIndex(idx2)
+			}
+		}
+	}
+
+	if idx2 < length2 {
+		key2 = ra2.getKeyAtIndex(idx2)
+		for key2 <= last {
```
lemire: This loop here looks a lot like one of the ``appendCopiesUntil`` functions in ``roaringArray``. Would you either use of these functions or add a comment to explain why custom code is needed?

parallel.go
```diff
@@ -329,3 +331,289 @@ func ParAnd(parallelism int, bitmaps ...*Bitmap) *Bitmap {
...
+			idx2++
+			if idx2 == length2 {
+				break
+			}
+			key2 = ra2.getKeyAtIndex(idx2)
+		}
+	}
+
+	if idx1 < length1 {
+		key1 = ra1.getKeyAtIndex(idx1)
```
lemire: Same comment than for above loop.

parallel.go
```diff
@@ -329,3 +331,289 @@ func ParAnd(parallelism int, bitmaps ...*Bitmap) *Bitmap {
...
+
+				key1 = ra1.getKeyAtIndex(idx1)
+				key2 = ra2.getKeyAtIndex(idx2)
+			}
+		}
+	}
+
+	if idx2 < length2 {
+		key2 = ra2.getKeyAtIndex(idx2)
+		for key2 <= last {
```
lemire: Same comment, make sure that an append function that does the same thing does not already exist in our code base.

parallel.go
maciej: To be honest I was not aware of `appendCopiesUntil`. I'll check it out!

lemire: @maciej Right. So it is just a case of software engineering hygiene. If we already have the functions, let us see if they can be reused. Nothing wrong with rolling your own, but it may increase the line count unnecessarily.

parallel.go
maciej: :+1: I'll fix that. There is no reason for the branches!

lemire:
@maciej Let me know when you recommend merging.

maciej:
@lemire I've addressed your points from the review. As for `appendCopiesUntil` – you can only specify the stoppingKey without the start index. I decided not to introduce another method to the _roaringArray_ struct receiver. I think using the general `appendCopy` which abstract the CoW logic away is a better solution.

That said, once the CI checks are done I recommend merging.

lemire:
The checks have failed, but it might be bogus. Restarting.

lemire:
We are good now.

lemire:
@cakey Care to comment on the speed-up of the new code on your internal tests?

lemire:
This is now Version 0.4.5.

Send pizza to @maciej

PR #180, opened 2018-04-21
add a fail input data based on fuzz

fzerorubigd:
when the numRuns is high enough, 2 * numRuns could pass the uint16 limit


lemire:
Good catch. We know from first principles that 2*numRuns must be smaller than (1<<16) so if there is an overflow, it is clearly an error.

That is, if numRuns is (1<<15) then the run container will exceed 8kB, so it is not efficient, and it should not occur in real life.

Can we do something like the following...?

```
if numRuns > 0x7fff {
  // bail out
}
```

fzerorubigd:
There is a problem with `1<<16` , the next line is `by = make([]byte, numRuns * 4)` so basically the problem still persist for `by` too.

lemire:
@fzerorubigd

Ok. Merging.

lemire:
If your name isn't in the CONTRIBUTORS file please add it though a PR.

PR #182, opened 2018-04-22
push coverage report to travis (again)

fzerorubigd:
also add my name to CONTRIBUTORS file as requested.

based on my experience the COVERALLS_TOKEN is needed, but based on goveralls readme : 

> For a public github repository, it is not necessary to define your repository key (COVERALLS_TOKEN).

so, try it without token first :) 
also if the goveralls fails, then try the normal `go test` like before. 
another thing, the repo token was publicly available in old `.travis.yaml` , it is advised to keep it a secret, so regenerating a new token is recommended.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,17 +14,12 @@ go:
 - 1.10.x
 - tip
 
-# Next lines seem to fail horribly
-#- go test -v -covermode=count -coverprofile=coverage.out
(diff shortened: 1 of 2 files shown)


fzerorubigd:
Also, rle.go is not generated (wild guess :) based on file name), and there is lot of unused code in that file. why ignoring it?

lemire:
The file rle.go is not used directly, but mapped to a file we do use which gets tested, see our Makefile. It is kind of odd but it comes about because Go lacks generics. So rle.go is our poor's man generic.

It could actually be removed from the repo. It is kept for historical and documentation reasons.

fzerorubigd:
@lemire is there any problem with this PR? i think it's ready to merge :)

lemire:
Sure, I don't think there is a problem, only lack of time. Merging.

Issue #185, opened 2018-05-23
allow backward iteration (descending order)


lemire:
One should be able to iterate from the end to the beginning... in descending order...

This is implemented in the [Java version of Roaring](https://github.com/RoaringBitmap/RoaringBitmap).

http://static.javadoc.io/org.roaringbitmap/RoaringBitmap/0.7.10/org/roaringbitmap/RoaringBitmap.html#getReverseIntIterator--

> a custom iterator over set bits, the bits are traversed in descending sorted order

lemire:
Implementation idea.

We have three container types... ``bitmapcontainer.go``, ``arraycontainer.go`` and ``rlei.go``. They currently implement ``shortIterable``:

```
getShortIterator() shortIterable
```

```
// Iterator creates a new IntIterable to iterate over the integers contained in the bitmap, in sorted order
func (rb *Bitmap) Iterator() IntIterable {
	return newIntIterator(rb)
}


type intIterator struct {
	pos              int
	hs               uint32
	iter             shortIterable
	highlowcontainer *roaringArray
}

// HasNext returns true if there are more integers to iterate over
func (ii *intIterator) HasNext() bool {
	return ii.pos < ii.highlowcontainer.size()
}

func (ii *intIterator) init() {
	if ii.highlowcontainer.size() > ii.pos {
		ii.iter = ii.highlowcontainer.getContainerAtIndex(ii.pos).getShortIterator()
		ii.hs = uint32(ii.highlowcontainer.getKeyAtIndex(ii.pos)) << 16
	}
}

// Next returns the next integer
func (ii *intIterator) Next() uint32 {
	x := uint32(ii.iter.next()) | ii.hs
	if !ii.iter.hasNext() {
		ii.pos = ii.pos + 1
		ii.init()
	}
	return x
}

func newIntIterator(a *Bitmap) *intIterator {
	p := new(intIterator)
	p.pos = 0
	p.highlowcontainer = &a.highlowcontainer
	p.init()
	return p
}
```

This looks like a lot of code, but it is straight-forward if you realize that roaring is just an array of containers.


So we probably want to implement something like this at the container level...

```
getReversedShortIterator() shortReversedIterable
```

And then this needs to be implemented for all three container types.


For reasons that are hard to explain, and probably bad, ``rlei.go`` relies on ``rle16.go`` which itself has been originally automatically transcoded from ``rle.go``. We don't actually need ``rle.go``.

lemire:
[...]

PR #189, opened 2018-05-29
Add Reverse Iterators - resolves issue #185 allow backward iteration …

joenall:
Added reverse iterators. PrevSetBit is not using math/bits and may be improvable. Test coverage is good and all tests pass. Feedback on names and style are welcome.

### DIFF
--- a/arraycontainer.go
+++ b/arraycontainer.go
@@ -28,6 +28,10 @@ func (ac *arrayContainer) getShortIterator() shortIterable {
 	return &shortIterator{ac.content, 0}
 }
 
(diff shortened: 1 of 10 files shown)


maciej:
looks got to me

lemire:
@maciej's recommendation is good to have. Let me see.

lemire:
It is "improvable" but a great basis. Merging.

lemire:
I only made a small change (regarding the bitset containers) and then released it. Basically, Joe did all the hard work.

PR #195, opened 2018-06-23
Add FlipMaxRangeEnd tests and MaxRange

anacrolix:
TestPackageFlipMaxRangeEnd panics, and TestBitmapFlipMaxRangeEnd goes into an infinite loop. MaxRange is a desperately needed constant.


lemire:
This seems to only add tests, so safe enough to merge.

lemire:
Bug fixed in latest release.

PR #196, opened 2018-06-30
remove unnecessary clone from arrayContainer.iand()

cakey:
I believe this clone is unnecessary for the in-place intersection. Showed up on our profiling!


cakey:
Looks like that branch for missing a test according to the coverage

lemire:
Yeah. It is clearly a performance bug. (Not a logical one though.)

Excellent.

Issue #163, opened 2018-03-30
WriteTo() optimization ideas?


steveyen:
It appears the current implementation of Bitmap.WriteTo() isn't very much different than the user instead calling ToBytes() and invoking io.Writer.Write() itself...
  https://github.com/RoaringBitmap/roaring/blob/master/roaringarray.go#L526

Was wondering if there are approaches or ideas that might help avoid the "extra" toBytes() allocation?

I'd be willing to give an improvement an attempt if it's pretty straightforward.  In my usage of roaring in a golang full-text engine (blevesearch), the io.Writer that we're using is one of those bufio buffered writers, so the guess would be that a bunch of Write()'s ought to be ok.

cheers,
steve

lemire:
The bulk of the time in the ``toBytes`` function is going to be spent there:

```
	stream := &bytes.Buffer{}
        (...)
	for i, c := range ra.containers {
		_ = i
		_, err := c.writeTo(stream)
		if err != nil {
			return nil, err
		}
	}
```


https://github.com/RoaringBitmap/roaring/blob/master/roaringarray.go#L512-L518


(Of course, one should do actual profiling... results could vary depending on your application.)

I don't think it is hard to pull out this code from the ``toBytes`` function and do something more direct with it.

It is less obvious that it is a good path to speed things up. Buffering is not necessarily a negative performance wise. It can be an optimization too! 

A better strategy might be to make it possible to reuse the same memory buffer several times. Or something.

Anyhow, writing the code is not hard (I expect you will agree0. What is hard is to benchmark and tune.

steveyen:
> A better strategy might be to make it possible to reuse the same memory buffer several times.

Indeed -- that's the pattern we've been following in blevesearch, and it's been quite a beneficial trick to play again and again.

> What is hard is to benchmark and tune.

So true!!

lemire:
I marked this "help wanted". I think that this is an accessible problem for contributions.

PR #197, opened 2018-07-09
Reduce buffering in WriteTo

bpot:
This greatly reduces allocations when the writer has enough space to
hold the serialized bitmap.

```
name                   old time/op    new time/op    delta
SerializationSparse-8     127µs ± 6%      50µs ± 4%  -60.17%  (p=0.008 n=5+5)
SerializationMid-8       54.4µs ± 4%     8.2µs ± 4%  -84.98%  (p=0.008 n=5+5)
SerializationDense-8     6.43µs ± 4%    0.66µs ±13%  -89.78%  (p=0.008 n=5+5)

name                   old alloc/op   new alloc/op   delta
SerializationSparse-8     410kB ± 0%      12kB ± 0%  -96.97%  (p=0.008 n=5+5)
SerializationMid-8        277kB ± 0%       1kB ± 0%  -99.50%  (p=0.008 n=5+5)
SerializationDense-8     37.0kB ± 0%     0.1kB ± 0%  -99.61%  (p=0.008 n=5+5)

name                   old allocs/op  new allocs/op  delta
SerializationSparse-8      8.00 ± 0%      2.00 ± 0%  -75.00%  (p=0.008 n=5+5)
SerializationMid-8         10.0 ± 0%       2.0 ± 0%  -80.00%  (p=0.008 n=5+5)
SerializationDense-8       5.00 ± 0%      2.00 ± 0%  -60.00%  (p=0.008 n=5+5)
```

#163 points out that buffering may be preferred to calling `Write` on the underlying `io.Writer` many times but I think this change provides the most flexibility. It allows a user to provide a `bytes.Buffer` or `bufio.Writer` as the `io.Writer` if they want additional buffering.

Our use-case is that we're hashing the serialized representation of the bitmap but not actually storing the serialized representation in most cases -- we don't want the serialized representation buffered at all.

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -469,6 +469,7 @@ func BenchmarkSparseIterateBitset(b *testing.B) {
 }
 
 func BenchmarkSerializationSparse(b *testing.B) {
+	b.ReportAllocs()
 	b.StopTimer()
 	r := rand.New(rand.NewSource(0))
 	s := NewBitmap()
@@ -487,6 +488,7 @@ func BenchmarkSerializationSparse(b *testing.B) {
 }
 
 func BenchmarkSerializationMid(b *testing.B) {
+	b.ReportAllocs()
 	b.StopTimer()
 	r := rand.New(rand.NewSource(0))
 	s := NewBitmap()
@@ -505,6 +507,7 @@ func BenchmarkSerializationMid(b *testing.B) {
 }
 
 func BenchmarkSerializationDense(b *testing.B) {
+	b.ReportAllocs()
 	b.StopTimer()
 	r := rand.New(rand.NewSource(0))
 	s := NewBitmap()
--- a/roaringarray.go
+++ b/roaringarray.go
@@ -457,8 +457,7 @@ func (ra *roaringArray) serializedSizeInBytes() uint64 {
 //
 // spec: https://github.com/RoaringBitmap/RoaringFormatSpec
 //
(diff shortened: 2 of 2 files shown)


lemire:
This seems good. Merging.

I do have a question however... isn't there a faster way to hash than to serialize/hash. That seems inefficient.

bpot:
I was wondering the same thing. Although with this change directly hashing the containers as byte buffers doesn't seem too inefficient. I haven't benchmarked it but it seems better than converting to an array or iterating.

lemire:
@bpot When the output fits in cache, it is probably ok. You don’t have large bitmaps?

bpot:
@lemire I guess it depends on what you mean by large. Almost all will fit in a single core's cpu cache on modern processors.

lemire:
@bpot 

> Almost all will fit in a single core's cpu cache on modern processors.

... which is a fine definition for "small". 

Still, it would not be hard to be 2x or more faster... if this is something that is done regularly in your application, it might be worth it to invest in a fast function that we could merge in the main lib.

bpot:
@lemire 

A `Checksum` method that was dependent on the logical contents of the bitmap (i.e. was the same regardless of run optimization) would be useful.

bpot:
@lemire any changes I need to make before I can get this merged?

lemire:
Sorry. I got distracted by the side question.

PR #200, opened 2018-08-01
AndNot: avoid returning bitmapcontainer with arraycontainer cardinality

bpot:
In some cases this could cause bitmaps to be serialized incorrectly.

Check if a bitmap needs to be converted to an array container after an
inplace-andNot operation.

This also adds a sanity check when writing bitmap containers to make
sure the cardinality is correct.

### DIFF
--- a/serialization_test.go
+++ b/serialization_test.go
@@ -493,12 +493,11 @@ func TestSerializationRunOnly033(t *testing.T) {
 func TestSerializationBitmapOnly034(t *testing.T) {
 
 	Convey("bitmapContainer writeTo and readFrom should return logically equivalent containers", t, func() {
-
 		seed := int64(42)
 		rand.Seed(seed)
 
 		trials := []trial{
-			{n: 1010, percentFill: .50, ntrial: 10},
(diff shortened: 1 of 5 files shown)


serialization_test.go
```diff
@@ -493,12 +493,11 @@ func TestSerializationRunOnly033(t *testing.T) {
 func TestSerializationBitmapOnly034(t *testing.T) {
 
 	Convey("bitmapContainer writeTo and readFrom should return logically equivalent containers", t, func() {
-
 		seed := int64(42)
 		rand.Seed(seed)
 
 		trials := []trial{
-			{n: 1010, percentFill: .50, ntrial: 10},
+			{n: 8192, percentFill: .99, ntrial: 10},
```
bpot: This had to be updated to set more bits or it would trip the new check in `writeTo`.

bpot:
We could also validate that array containers have strictly increasing values which, I think, would have caught this on the deserialization side. I'm not sure if we want to be checking that every time we deserialize a bitmap though. I suppose we could add a `Verify` method of some sort.

lemire:
Very good.

PR #210, opened 2018-12-10
add fail test, removing item, invalidate the iterator

fzerorubigd:
in large number, it may panic, but in lower numbers, it makes the hasNext function
invalid.

Also, I am not sure if removing an index in the iterator is a good thing to do.

I am not sure about a fix yet. so just sharing the problem.


lemire:
Hmmm... I am uneasy about merging this because I do not expect the test you present to pass... Maybe this is a failure in our documentation, but the assumption with an iterator is that you are not modifying the underlying container as it invalidates the iterator.

fzerorubigd:
I am not suggesting to merge this. It is easier to write a test instead of opening an issue for me. So this is just an issue. If this is intended behavior, then adding it to the documents is the fix, if not, then we should fix it somehow.

PR #263, opened 2020-07-25
Introduce an alternative backing implementation for Roaring Bitmaps.

jacksonrnewhouse:
I've spent the last week implementing an alternative backing store for Roaring, with the goal of minimizing allocation calls and garbage collection. I develop a system that stores a large amount of immutable data on disk, and being able to quickly read, intersect and count the underlying bitmaps is of critical importance. Profiling has shown that a significant amount of CPU (often over 30%) was being spent on allocation tasks. Additionally, when garbage collection ran, it imposed significant performance penalties on our overall latency.

This work follows on some experimentation I did with combining the scanning of a []byte with Oring it in-place into another segment. [FastByteOr](https://github.com/jacksonrnewhouse/roaring/blob/inline_or/roaring.go#L1005) was a hacky attempt at this, but did show a 30% improvement in wallclock and overall CPU and about halved allocations. However, it required that you really only needed a single-pass, and generally would hurt maintainability.

Rather than continuing that path, I implemented a Roaring library that is primarily backed by a single []slice, plus possibly a header slice and a couple of ints. The basic structure supports reading from no-run bitmap files (SERIAL_COOKIE_NO_RUNCONTAINER), performing just a quick check that the slice is long enough. This bitmap is assumed to e immutable, which isn't terribly restrictive given that almost any update would require shifting a large percentage of the data.

For supporting mutability of bitmaps, particularly through set operations, we introduce a different format. The containers is the number of nonempty containers, while maxContainers is a bookkeeping tool to ensure that we don't have to repeatedly copy data between slices. Header looks just like bytes [8, 8 + 4 * containers) in the standard format. Finally, for writable bitmaps the data for container with key s is stored in a 8192 byte range, starting at s * 8192. While this almost surely increases the memory footprint of live bitmaps, it makes the data reliably addressible and minimizes copies. Because these bitmaps will typically occupy rather goodsized segments, the Clear() and Close() methods provide some nice methods to allow for reuse. Clear() should be used when the active gothread will be able to immediately reuse it, while Close() releases the backing slice to a sync.Pool in order to be recycled.

This is a pretty giant file, as it needs to lift a dozen algorithms from various parts of the repo. That being said, we were able to see significant performance improvements. We ran two queries twice, and saw the following results with the change:

Query 1
Prod: 
Malloc: 10.3 billion
TotalAlloc: 637 billion
total CPU: 13,500 units
runtimes: 28s, 28s

Test:
Malloc: 500 million (95% reduction)
TotalAlloc: 51 billion (92% reduction)
total CPU: 11,500 units (15% reduction)
runtimes: 23s, 23s (18% reduction)

Query 2
Prod:
Malloc: 52 billion
TotalAlloc: 3 trillion
total CPU: 66,350
runtimes:  135s, 132s

Test:
mallocs: 4.36 billion (91% reduction)
TotalAlloc: 410 billion (86% reduction)
total CPU: 23,000 units (65% reduction)
runtimes: 66s, 60s (50% reduction)

This is despite several shortcuts and inefficiencies in the new code. For instance, I have yet to implement galloping AND cardinality in the new format, and it currently spends 45% of one query doing union2by2Cardinality.

I'm curious if people think this is an avenue worth introducing into the mainline. If not, it is probably that we'll adopt some version of this, as the performance benefits are considerable.

### DIFF
--- a/lowalloc.go
+++ b/lowalloc.go
@@ -1,0 +1,1555 @@
+package roaring
+
+import (
+	"fmt"
+	"math/bits"
+	"sync"
+	"unsafe"
+)
+
+const (
+	containerStorageBytes     = 8192
+	bitmapLongCount           = containerStorageBytes >> 3
+	containerCountOffset      = uint32(4)
+	containerDescriptionStart = uint32(8)
+	bytesPerContainer         = uint32(4)
+
+	maxContainers          = uint32(256)
+	fatWritableSliceLength = int(maxContainers * containerStorageBytes)
+
+	// 0: 2 bytes for key,
+	// 2: 2 bytes for cardinality - 1
+	cardinalityIncrement = uint32(2)
+)
+
+var writableArrayPool = sync.Pool{
+	New: func() interface{} {
+		return make([]byte, 0, fatWritableSliceLength)
+	},
+}
+
+type RoaringBitmap struct {
+	data         []byte
+	header       []byte
+	containerMax uint32
+	containers   uint32
+	writable     bool
+}
+
+func (bitmap *RoaringBitmap) FromBuffer(bytes []byte) error {
+	cookie := ReadSingleInt(bytes, 0)
+	if cookie != serialCookieNoRunContainer {
+		return fmt.Errorf("unexpected signature %d", cookie)
+	}
+	size := ReadSingleInt(bytes, 4)
+	if size > 0 {
+		pointer := 8 + (4 * (size - 1))
+		finalContainerCardinality := ReadSingleShort(bytes, pointer+2)
+		pointer += 4 * size
+		finalOffset := ReadSingleInt(bytes, pointer) + uint32(lengthFromCardinalityShort(finalContainerCardinality))
+		if finalOffset != uint32(len(bytes)) {
+			return fmt.Errorf("expect the bitmap to be %d bytes, not %d bytes",
+				finalOffset,
+				len(bytes))
+		}
+	}
+	bitmap.data = bytes
+	bitmap.writable = false
+	return nil
+}
+
+// Contains returns true if the integer is contained in the bitmap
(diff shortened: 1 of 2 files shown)


lemire:
This seems very promising.

>  performing just a quick check that the slice is wrong enough

Is this a typo...?

> For supporting mutability of bitmaps, particularly through set operations, we introduce a different format

Is that meant to be mutability on disk, or mutability in memory?

jacksonrnewhouse:
yeah, should've been "long enough". 

I mainly meant mutability in memory. The use case I am targeting is reading bitmaps off of disk at query time, composing them via set operations, and then calling Cardinality() and AndCardinality(), after which the bitmaps can be released.

guymolinari:
I'm doing something similar but I hold on to the bitmaps in memory in a
large cache.  Mutability is a requirement for me.  Performance improvements
of >50% are interesting though.

foubarre:
That's very interesting, so I can read a bitmap from disk allocation free if I mmap it and pass it to your immutable version of FromBuffer()?

lemire:
Note that this would need a lot more testing, benchmarking and validation before we could merge this. If folks are interested, please don't be shy about contributing.

jacksonrnewhouse:
I'm going to go ahead and close this for now. I have a more workable ImmutableBitmap approach that I may try to work into a future PR.

Oppen:
Any news on the new version? Otherwise, I think I **may** find the time to test it in some of my work's datasets. It's a long enough set of bitmaps that I have to tar them together lest I want `ls` to get a reasonably beefy machine freeze for minutes. I just load those and `or` them together in a single bitmap. I have about 185 of those tars alive to test.

I don't think it's usable officially for my work because the input files come from a different team, but I may make the case for it if I have a working implementation upstream.

On another note, is there any measure of how this affects on-disk size? How about after compression?

PR #290, opened 2020-11-02
Modified the implementation of ReadFrom to support roaring 64 to directly read roaring 32 data.

davidchen-cn:
Hi reviewer, I think this can help roaring32 users to smoothly migrate to roaring64.


guymolinari:
Good feature.  +1

roaring.go
```diff
@@ -67,11 +67,11 @@ func (rb *Bitmap) WriteToMsgpack(stream io.Writer) (int64, error) {
 // The format is compatible with other RoaringBitmap
 // implementations (Java, C) and is documented here:
 // https://github.com/RoaringBitmap/RoaringFormatSpec
-func (rb *Bitmap) ReadFrom(reader io.Reader) (p int64, err error) {
+func (rb *Bitmap) ReadFrom(reader io.Reader, cookieHeader ...byte) (p int64, err error) {
```
lemire, line 70: Could you document the new parameter `cookieHeader ...byte`?

lemire:
@guymolinari Would you review it? I just looked at it superficially. It might be useful to have someone think it through.

guymolinari:
Will do.

jacksonrnewhouse:
I have no concerns with the general idea here, but it does seem to ugly-up some methods in ways I'd typically rather avoid. For instance, `func (ra *roaringArray) readFrom(stream byteInput, cookieHeader ...byte) (int64, error)` is a much uglier method signature than `func (ra *roaringArray) readFrom(stream byteInput) (int64, error)`.

From what I can read, it basically is injecting some values from `cookieHeader` that would normally be read from the `byteInput`. How possible would it be to instead factor out all of the reading below the changes, and then have two separate pathways. Something like `func (ra *roaringArray) readWithhHeaders(stream byteInput, format formatType)` where `formatType` is a type encompassing the possible headers. That way we could avoid having logic in 32 bit roaring that is just there to support 64 bit roaring.

guymolinari:
All good thoughts.    Stepping back, I would suggest that we also consider
compatibility with the other Roaring library implementations.   This is a
stated capability in that the wire protocol is specifically supposed to be
cross platform.   I have concerns about the existing 64 bit implementation
not being compatible with the C/C++ 64 bit implementation.   I think we
should have specific test cases verifying that we can seamlessly
interoperate.

Sorry for broadening the scope on this folks but I really think we should
do this the right way.

Guy

lemire:
Broadening the scope should be done with care in the sense that we should not ask of a single PR what it cannot provide.

64-bit is a bit of a mess in the sense that there is *no* cross-language compatibility (at all). You have one implementation in C++, *two* implementations in Java (both in the RoaringBitmap library), you have one in roaring. People have said good things about the more recent Java 64-bit implementation and it is definitively being used in production.

What one would need to do is to sit down, look and them, and propose a standard format that everyone could agree to. It is not hard as in "curing cancer", but it requires a deliberate effort.

guymolinari:
I can test this out as is then.  Sometimes good is the enemy of perfect.

lemire:
@guymolinari  Another thoughts regarding this PR? I think we need to either accept it, or make some concrete recommendations to @davidchen-cn   I think that @jacksonrnewhouse made some good points about how this PR does make things "uglier" at times.  But I think we are a bit too vague about the fix...

AskAlexSharov:
@guymolinari , hello. do you have any other comments for this PR? 
Theoretically can move all "container" types from 32 and 64 packages to "internal" package, but it maybe a big move.

guymolinari:
I would just merge it.  We could refactor later if needed.

lemire:
@davidchen-cn Would you sync with our main branch? Sorry about this, but it seems that there were changes and the two versions now differ.

lemire:
@davidchen-cn  Ping! Would you sync your fork?

davidchen-cn:
@lemire Sorry for reply late. I will sync it soon.

lemire:
Merging. To be revised later.

PR #351, opened 2022-01-16
Use testing facilities for logging on error

Oppen:
- Reduce arbitrary printing during tests
- Remove helper for tests from application builds

Examples remain printing as an exception, as they're intended to be seen
by users learning the package.


lemire:
All good. Let us merge.

guymolinari:
Are the API changes also applicable to roaring 64?

Oppen:
This shouldn't affect any API. It only changes tests.

PR #382, opened 2023-02-24
Modify BSI existence bitmap on Increment

anacrolix:
Somehow this got missed when BSI was added. There are no existing tests that check this directly, but it's easily discoverable.


guymolinari:
Good catch Matt!

lemire:
Merging.

PR #383, opened 2023-03-03
BSI fixes

anacrolix:
Some improvements from extensive use under load. More details in the commit messages.

The concurrency is interesting, I suspect that everything was made concurrent to start with, and I think some larger operations might benefit (like And and Or), but some do not.

### DIFF
--- a/BitSliceIndexing/bsi.go
+++ b/BitSliceIndexing/bsi.go
@@ -2,11 +2,12 @@ package roaring
 
 import (
 	"fmt"
-	"github.com/RoaringBitmap/roaring"
 	"math/bits"
 	"runtime"
 	"sync"
 	"sync/atomic"
+
+	"github.com/RoaringBitmap/roaring"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


guymolinari:
I'm fine with simplification,  however I would ask for performance
benchmarking to confirm that performance (latency) is not impacted.  I'm
storing 64 bit values and run on multi-core platforms

anacrolix:
No worries, the Runoptimize change fixes a bug if you want to cherry pick just that.

lemire:
@guymolinari Can you suggest a before-after benchmark we could to test out performance ?

guymolinari:
I routinely store highly distributed 64 bit values (think hashes).   I
would simply store a few million 64 bit hashes in a BSI then measure the
"add" operation latencies on a single core vs. multi-core

Guy

lemire:
@anacrolix Can you just run a basic benchmark to see what effect your PR has?

lemire:
@guymolinari  Any chance you could benchmark it? Before and after, based on your use case?

guymolinari:
Yes.   Will do ASAP.

anacrolix:
> No worries, the Runoptimize change fixes a bug if you want to cherry pick just that.

If it wasn't clear, this commit is an unrelated bug and could be cherry picked independently.

lemire:
I am sure @guymolinari will get back to us soon?

anacrolix:
Here's the scheduler overhead from a 5s snapshot with a parallelized SetValue (i.e. more than 100% CPU for the scheduler overhead): 
image
After the fix I've provided this is the overhead
image
(i.e. negligible items elsewhere).
It might make sense to parallelize SetValue but it would require chunking across the bitmaps, and there would need to be a lot of them. I think for 99% of users, a sequential SetValue is going to be much much faster. My own program had a 3-4x speedup just by removing this pathological scheduler case.

I've been very conservative and only fixed this one case that performs very badly. I haven't seen an issue elsewhere.

anacrolix:
I've been using this extensively in production for a search database implementation. In particular I did testing around whether removing RunOptimize was correct, and whether this is faster. I am certain in both cases. I strongly recommend to merge this.

lemire:
Merged.

Smallhi:
I implemented BSI in C. for some reason, we didn't use concurrency, It seems the performance is OK. (postgres-like env)

guymolinari:
BTW, this is good stuff.  I'm glad you are all digging into this deeply.   I did approve the PR earlier but didn't comment further at the time.   I've been using this as part of a DB engine that is very close to being a drop in replacement for MySQL and will use the released version of this in my project.

anacrolix:
> Also, one final thought WRT parallelization. If we could migrate all of roaring bitmap to run on GPUs, would multithreading this matter?

No, and I think from the C version, multithreading is only worthwhile in very specific cases.

PR #451, opened 2024-10-23
BSI library expanded to include support for large arbitrary values via the big.Int library

guymolinari:
There is no fundamental reason why the BSI library cannot support values exceeding 64 bits.  Why is this necessary?   It would  now be possible to store values like UUID and 96 bit timestamps, as well as advanced string encodings in a BSI.

The BSI library has been modified to preserve the existing int64 interface and has been expanded to leverage the golang big.Int math library.   The original int64 methods now convert incoming values to *big.Int (and vice versa, etc).

Additional test coverage has been added and some of the original tests have been expanded.   

Performance does not seem to be substantially impacted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/RoaringBitmap/roaring/v2
 
-go 1.22
+go 1.15
 
 require (
 	github.com/bits-and-blooms/bitset v1.12.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/RoaringBitmap/roaring/v2
 
-go 1.22.0
+go 1.22
 
 toolchain go1.23.2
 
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,9 +1,18 @@
 module github.com/RoaringBitmap/roaring/v2
 
-go 1.14
+go 1.22.0
```
lemire, line 3: Some tests in CI complain about the syntax here.

lemire:
@guymolinari Can you have a look at our failing tests?

guymolinari:
Looking.....

guymolinari:
Daniel.   Would it be possible to update the environment to use at least go 1.15?

lemire:
> Daniel. Would it be possible to update the environment to use at least go 1.15?

Absolutely. Maybe it would be cleaner to do it with a separate PR.

We could push the required version up, and then handle this PR.

guymolinari:
Daniel.   Missed your last message.   I think I have it sorted out now.

lemire:
Let us merge.

Issue #455, opened 2024-10-29
MinMaxBig may set default foundSet?


ChicChip:
v2.4.0
in "func (b *BSI) MinMaxBig(parallelism int, op Operation, foundSet *Bitmap) *big.Int"
before excute  "card := foundSet.GetCardinality()" need add code
"
if foundSet == nil {
		foundSet = &b.eBM
}
"
then when foundSet is nil , this function will return b.eBM's max or min value?

guymolinari:
The purpose of a 'foundSet' is to restrict the operation to a subset of the
contents of a BSI (or other operations such as comparisons).   If a nil
foundSet is passed
then the entire BSI is considered.   If you look at the code you will see
that if a foundSet is not provided, then the existence bitmap is used
instead.  The existence
bitmap of a BSI indicates that a value is present (non NULL) for a given
columnId.

guymolinari:
This was a question for clarification and not a bug or feature request so I am closing the issue.

ChicChip:
it didn't use existence bitmap in **MinMaxBig** func, and when i use 
`bsi.MinMaxBig(1,roaring64.MAX, nil)`
it will throw a panic
`github.com/RoaringBitmap/roaring/v2/roaring64.(*Bitmap).GetCardinality(0xc000191308?)
        /Users/jidian.shan/Desktop/user-profile-be/vendor/github.com/RoaringBitmap/roaring/v2/roaring64/roaring64.go:390 +0x18
github.com/RoaringBitmap/roaring/v2/roaring64.(*BSI).MinMaxBig(0xc0004d6000, 0x4e1f?, 0x8, 0x152f2e2?)
        /Users/jidian.shan/Desktop/user-profile-be/vendor/github.com/RoaringBitmap/roaring/v2/roaring64/bsi64.go:512 +0x70
`

guymolinari:
Yes.   Indeed this should be fixed.  There are a few other places where we should check nil as well.   However, this works per the unit tests.

func TestMinMaxBigWithNegative(t *testing.T) {
    bsi := setupAutoSizeNegativeBoundary()
    assert.Equal(t, big.NewInt(-5), bsi.MinMaxBig(0, MIN, bsi.GetExistenceBitmap())).       //WORKS
    assert.Equal(t, big.NewInt(5), bsi.MinMaxBig(0, MAX, bsi.GetExistenceBitmap()))         //WORKS
[...]

PR #456, opened 2024-10-31
fix use nil parameter, the MinMaxBig func will throw panic's problem

ChicChip:
i fixed fix use nil parameter, the MinMaxBig func will throw panic's problem (https://github.com/RoaringBitmap/roaring/issues/455)

### DIFF
--- a/BitSliceIndexing/bsi.go
+++ b/BitSliceIndexing/bsi.go
@@ -435,6 +435,10 @@ func (b *BSI) MinMax(parallelism int, op Operation, foundSet *roaring.Bitmap) in
 
 	resultsChan := make(chan int64, n)
 
+	if foundSet == nil {
+		foundSet = b.eBM
+	}
+
 	card := foundSet.GetCardinality()
 	x := card / uint64(n)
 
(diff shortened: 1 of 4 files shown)


guymolinari:
Could you also look at CompareBigValue ad well?

ChicChip:
in bsi64.go and bsi.go file, if foundSet is nil, is it all set to exisitence bitmap?

ChicChip:
i think these function may all hava nil pointer problem when parameter is nil

1. TransposeWithCounts
2. SumBigValues
3. IntersectAndTranspose
4. TransposeWithCounts
5. Increment

guymolinari:
Yes.  Also in CompareBigValue.  If 'end' is not set it will panic also

ChicChip:
> i think these function may all hava nil pointer problem when parameter is nil
> 
> 1. TransposeWithCounts
> 2. SumBigValues
> 3. IntersectAndTranspose
> --4. TransposeWithCounts
> 5. Increment

I have set default value for these function, include CompareBigValue

PR #473, opened 2025-04-08
`ToArray` with pre-allocated slices: `ToExistingArray`

TheHippo:
I have a use case were I need to call `ToArray` quite often. Every call creates a new slices which creates a lot of pressure for Go's GC. This PR adds a new function - `ToExistingArray` - to `Bitmap` which allows using a pre-allocated slice, that later can be re-used.


d-enk:
As if it were possible to do without a pointer to a slice 🤔

PR #521, opened 2026-04-10
fix: CardinalityInRange off-by-one for gaps between runs within a container

### LINKED ISSUE #520: CardinalityInRange returns wrong count for gaps between runs within a single container
`CardinalityInRange` returns a wrong count when the query range falls in a gap between two run-encoded intervals in the same container.

```go
rb := roaring.NewBitmap()
rb.AddRange(0, 1024)
rb.AddRange(2048, 3072)
rb.RunOptimize()

fmt.Println(rb.CardinalityInRange(1024, 2048)) // prints 1024, should be 0
```

Off-by-one in `runContainer16.getCardinalityInRange` line 2546: `lastIdx < firstIdx-1` should be `lastIdx < firstIdx`. When the query lands between two intervals, `firstIdx = lastIdx + 1` and the check lets it fall through to count the next interval.

ValentaTomas:
Fixes #520.

`runContainer16.getCardinalityInRange` checks `lastIdx < firstIdx-1` but should check `lastIdx < firstIdx`. When the query range falls in a gap between two run intervals within the same container, `firstIdx = lastIdx + 1` and the guard doesn't trigger, so the next interval (outside the query) gets counted.

### DIFF
--- a/roaring_test.go
+++ b/roaring_test.go
@@ -1155,6 +1155,32 @@ func TestCardinalityInRangeRunOptimized(t *testing.T) {
 	assert.EqualValues(t, 3000+5000, rb.CardinalityInRange(2000, 75000))
 }
 
+// TestCardinalityInRangeRunGapWithinContainer tests CardinalityInRange when
+// two run intervals exist in the same container with a gap between them.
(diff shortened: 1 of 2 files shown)


lemire:
@DhruvilK7 Want to have a look?

PR #529, opened 2026-06-30
bitmap: use bits.LeadingZeros64 intrinsic in maximum()

TheHippo:
## Description

`bitmapContainer.maximum()` finds the highest set bit by counting the leading zeros of the top word. It used a hand-written helper, `clz()`, to do that counting. This change makes it use the standard library's `math/bits.LeadingZeros64` instead (already wrapped in this package as `countLeadingZeros`), and removes `clz()`.

## Type of Change

- [x] Performance improvement
- [x] Code refactoring
- [ ] Documentation update
- [ ] Test improvements
- [ ] Build/CI changes

### What was changed?

- `bitmapContainer.maximum()` now computes the highest set bit using  `countLeadingZeros(w)` (i.e. math/bits.LeadingZeros64`) instead of the  private `clz()` function.
- The hand-rolled `clz(i uint64) int` helper in `bitmapcontainer.go` is deleted (it had no other callers).

### Why was it changed?

`clz()` counted the leading zero bits of a word using a chain of `if` checks and bit shifts. The standard library already does the same thing with `math/bits.LeadingZeros64`, which the Go compiler turns into a single CPU instruction (`LZCNT` on amd64, `CLZ` on arm64). In short, `clz()` was a slower, handwritten version of something the standard library does in one instruction.

### How was it changed?

- Replaced `clz()` with `countLeadingZeros()`  the body of `maximum()`.
- Removed the `clz()` function.

## Testing

Passes. `maximum()` is exercised by the existing root-package tests (`roaring_test.go`) and the randomized `property_test.go`, which compare `Maximum()` against reference behavior over many bitmaps - these cover the substitution directly. No new test was required.

## Fuzzing

Not affected by this change (no serialization/parsing surface is touched), but the existing fuzz harness continues to build and run.

## Performance Impact

Negligible-to-positive.

### Performance Analysis

`benchstat` reports a 3.58% improvement over the entire benchmark suite (geomean). 

## Breaking Changes

None. Public API and behavior are unchanged.

### DIFF
--- a/bitmapcontainer.go
+++ b/bitmapcontainer.go
@@ -69,39 +69,11 @@ func (bc *bitmapContainer) safeMinimum() (uint16, error) {
 	return val, nil
 }
 
-// i should be non-zero
-func clz(i uint64) int {
-	n := 1
-	x := uint32(i >> 32)
-	if x == 0 {
-		n += 32
-		x = uint32(i)
-	}
-	if x>>16 == 0 {
-		n += 16
-		x = x << 16
-	}
-	if x>>24 == 0 {
-		n += 8
-		x = x << 8
-	}
-	if x>>28 == 0 {
-		n += 4
-		x = x << 4
-	}
-	if x>>30 == 0 {
-		n += 2
-		x = x << 2
-	}
-	return n - int(x>>31)
-}
-
 func (bc *bitmapContainer) maximum() uint16 {
 	for i := len(bc.bitmap); i > 0; i-- {
 		w := bc.bitmap[i-1]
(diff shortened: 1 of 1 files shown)


TheHippo:
Just for the record / clarification:

> the Go compiler turns into a single CPU instruction (`LZCNT` on amd64, `CLZ` on arm64)

* The arm64 optimization always happens
* The amd64 optimization only happens when compiled with `GOAMD64=v3`

lemire:
@TheHippo Yes. I know. Thanks for the PR.

PR #542, opened 2026-08-04
Optimize lazy bitmap container unions

perfloop-agent:
## Summary

Make bitmap-to-bitmap lazy unions write directly into a newly allocated result instead of cloning and immediately overwriting the receiver. The fixed-size container loops establish slice bounds once and process four words per iteration, removing repeated bounds checks while preserving deferred cardinality repair. This keeps the manual unrolling tied to the existing 1024-word bitmap-container invariant. Add a ParOr benchmark for four dense inputs with aligned bitmap containers and four workers.

## Performance

On workload `ParOr over four dense bitmaps with 64 aligned bitmap containers and four workers`, median metric `ns/op` changed from 370418 to 302755; paired median delta -56581 (-15.3% of baseline; at least 19/20 confidence interval -93983 to -41428 from 10 pairs).

## Testing

Ran the full Go test suite, the appengine build and test path, and 386, arm, and arm64 builds. Also checked simplified Go formatting and unnecessary conversions; the benchmark validates ParOr cardinality against a sequential union.

All 5 declared correctness checks passed.

Full verification record: https://app.perfloop.ai/t/oss/case_gh0svmf28c

### DIFF
--- a/bitmapcontainer.go
+++ b/bitmapcontainer.go
@@ -651,16 +651,42 @@ func (bc *bitmapContainer) lazyORArray(value2 *arrayContainer) container {
 
 func (bc *bitmapContainer) lazyIORBitmap(value2 *bitmapContainer) container {
 	answer := bc
-	for k := 0; k < len(answer.bitmap); k++ {
-		answer.bitmap[k] = bc.bitmap[k] | value2.bitmap[k]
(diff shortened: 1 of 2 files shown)


lemire:
Great optimization.

PR #546, opened 2026-08-12
Optimize AVX2 popcount inner loop and CPU feature detection

lemire:
David Sparks reviewed `popcnt_avx2_amd64.s` and sent a list of improvements
to it. This PR implements all of them; the credit is his.

## The main change

`VPSADBW` computes `|a-b|` per byte and sums each group of 8, so it can
absorb the per-byte add that `VPADDB` was doing. Feeding it the two nibble
counts directly gives `(B + lo) - (B - hi) = lo + hi`, which removes
`VPADDB` and its latency from the hot loop:

```diff
 #define COUNTBLOCK \
 	VPAND Ymask, Ydata, Ylo \
 	VPSRLW $4, Ydata, Yhi \
 	VPAND Ymask, Yhi, Yhi \
-	VPSHUFB Ylo, Ylut, Yc1 \
-	VPSHUFB Yhi, Ylut, Yc2 \
-	VPADDB Yc2, Yc1, Yc1 \
-	VPSADBW Yzero, Yc1, Yc1 \
+	VPSHUFB Ylo, Ylut1, Yc1 \
+	VPSHUFB Yhi, Ylut2, Yc2 \
+	VPSADBW Yc1, Yc2, Yc1 \
 	VPADDQ Yc1, Yacc, Yacc
```

This needs two lookup tables, one biased up by `B` and one subtracted from
`B`. The bias must satisfy `4 <= B <= 251` so that neither table wraps as
unsigned bytes (max nibble popcount is 4) and `a >= b` always holds, making
`VPSADBW`'s absolute value a no-op. `B = 15` is free, because `Ymask`
already holds 15 in every byte — so `SETUP` builds both tables with one
`VPSUBB` and one `VPADDB` and drops `Yzero` entirely.

## The rest

- Generic (VEX-encoded) AVX instructions accept an unaligned memory source,
  so the second input of the And/Or/Xor/Mask loops is read straight out of
  memory instead of being loaded into a register first. Only `VPANDN`'s
  *non-negated* operand may come from memory, so that loop loads `m` into
  the register and reads `s` from memory.
- `Ydata`/`Yhi`/`Yc2` and `Ylo`/`Yc1` have disjoint live ranges and now
  share registers: six architectural registers instead of ten.
- `lutmask` shrinks from 64 to 24 bytes. `VBROADCASTI128` duplicates the
  16-byte table into both 128-bit lanes and `VPBROADCASTB` splats the
  nibble mask from a single byte, so neither needs to be stored twice.
- `SHRQ` and `ANDQ` already set ZF, so the `TESTQ` instructions that
  followed them were redundant.
- `SETUP` moves below the branch that skips the vector loop entirely.
- The 64-bit tail loops use `POPCNTQ` with a memory source and a 32-bit
  loop counter (32-bit rather than 8-bit avoids a partial-register merge).
- `XORL` rather than `XORQ` for zeroing: no REX prefix, and some cores do
  not recognize `XORQ` as a zeroing idiom.
- `_hasAVX2` complements each feature word and `TEST`s it instead of
  `AND`/`CMP`, saving a large immediate per check, and all three checks now
  fall through to a single `SETEQ` that stores ZF into the bool result.

The inner loop goes from 13 instructions to 10. Assembly text shrinks from
987 to 914 bytes and the rodata blob from 64 to 24.

## Benchmarks

Intel Xeon Gold 6548N, go1.26.3, linux/amd64. Microbenchmarks, `count=10`:

```
                          │    base     │             new              │
PopcntSlice1024AVX2-128     205.0n ± 0%   180.0n ± 0%  -12.20% (p=0.000)
PopcntAndSlice1024AVX2-128  241.6n ± 1%   225.7n ± 1%   -6.58% (p=0.000)
Popcount-128                14.66n ± 1%   13.73n ± 1%   -6.35% (p=0.000)
PopcntSlice1024Go-128       305.2n ± 0%   305.1n ± 1%        ~ (p=0.811)
PopcntAndSlice1024Go-128    449.0n ± 1%   448.9n ± 0%        ~ (p=0.616)
```

The pure-Go fallbacks are untouched and measure flat, as a control.

Through the public API on dense bitmaps (32 bitmap containers), which is
what these routines actually serve, `count=8`:

```
                          │    base     │             new              │
DenseAndCardinality-128     9.404µ ± 1%   8.655µ ± 0%   -7.96% (p=0.000)
DenseOrCardinality-128      128.6µ ± 3%   120.8µ ± 4%   -6.02% (p=0.001)
```

## Testing

Full test suite passes on amd64. `TestAVX2PopcntDifferential` runs rather
than skipping, which confirms the rewritten `_hasAVX2` still returns true;
I also disassembled it and the mask loop to check the encodings landed as
intended. Beyond the existing differential test, I ran 20000 randomized
rounds over all five routines with unaligned sub-slices, random lengths,
and all-ones/all-zeros inputs — the numeric extremes of the bias trick —
all matching the Go reference.

### DIFF
--- a/popcnt_avx2_amd64.s
+++ b/popcnt_avx2_amd64.s
@@ -15,8 +15,7 @@
 // a byte is split into its low and high nibble, each nibble is looked up (one
 // VPSHUFB performs all 32 lookups in a 256-bit register at once), and VPSADBW
 // then combines the two results and sums each group of 8 byte-counts into a
-// 64-bit lane total that is accumulated (see COUNTBLOCK for how the two
-// nibble counts are added by VPSADBW itself rather than a separate VPADDB).
+// 64-bit lane total that is accumulated (COUNTBLOCK).
 // After the loop the four lane totals are summed (HSUM) into a scalar register.
 // Each iteration handles 256 bits (4 uint64); a scalar POPCNTQ tail handles
 // the trailing len%4 words, so any slice length is counted correctly.
@@ -39,8 +38,8 @@
 //     VZEROUPPER precedes every RET to avoid the AVX<->SSE transition penalty
 //     in any non-VEX SSE code that runs afterwards.
 
-// lutmask is a 24-byte read-only blob holding the two constants used by every
-// routine:
+// lutmask is a 17-byte read-only blob (the linker pads it out to whatever its
+// alignment requires) holding the two constants used by every routine:
 //   bytes  0..15 - the nibble popcount table, i.e. table[i] = number of set
 //                  bits in the 4-bit value i. Read low-byte-first, the first
 //                  qword 0x0302020102010100 is the bytes {0,1,1,2,1,2,2,3} for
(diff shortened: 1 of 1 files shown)


TheHippo (approved):
I'm the original author of the file and by no means an expert on this topic. The changes do look very reasonable, the tests pass and I can confirm the that the performance of the benchmarks has improved.

PR #556, opened 2026-08-25
Accelerate dense bitmap decoding in ToArray

perfloop-agent:
## Description

Speed up `Bitmap.ToArray` for dense bitmap containers on amd64 CPUs with AVX-512 support.

## Type of Change

- [x] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] Test improvements
- [ ] Build/CI changes

### What was changed?
- Added a density-gated AVX-512 decoder for dense bitmap containers.
- Added a `Bitmap.ToArray` benchmark covering four dense containers and allocation behavior.

### Why was it changed?
- Dense containers spend most of their decode time walking set bits with a scalar trailing-zero loop.
- Bulk decoding reduces that work without changing the ordered output or allocation contract.

### How was it changed?
- The decoder processes four ordered 16-bit output blocks at a time and advances the existing output cursor.
- Containers below the density threshold and targets without AVX-512 retain the existing scalar or generic path.

## Testing

- `go test ./...`
- `go build -tags appengine ./...`
- `go test -tags appengine ./...`
- Cross-builds for 386, arm, and arm64.
- `go tool unconvert ./...`
- Generated smat corpus and smat-hit tests.

Checks: 9 passed.

## Formatting

Ran `gofmt` and verified that no Go files require formatting changes.

## Fuzzing

Ran the generated smat corpus and smat-hit tests:

```bash
go test -tags=gofuzz -run='^TestGenerateSmatCorpus$' -count=1
go test -tags=gofuzz -run='^TestSmatHits$' -count=1
```

## Performance Impact

The fast path applies only to dense bitmap containers on supported amd64 CPUs. Sparse containers, appengine builds, and unsupported targets keep the existing fallback. The change keeps `ToArray` allocation and container traversal in their existing owners.

**Workload:** `ToArray over four dense bitmap containers`

| Metric | Before | After | Change |
| --- | --- | --- | --- |
| `ns/op` | 207344 | 94559 | 54.4% lower |

## Breaking Changes

None.

## Related Issues

None.

## Additional Notes

The output remains ordered and the portable implementation remains available as the fallback.

---

Generated by [Perfloop](https://app.perfloop.co). [Measurements and checks](https://app.perfloop.co/t/perfloop/case_0k4q7bje0n).


lemire:
Ths is promising. We use a similar approach in simdjson.

fillbits_avx512_amd64.go
```diff
@@ -0,0 +1,15 @@
...
+
+package roaring
+
+//go:noescape
+func fillLeastSignificant16bitsAVX512(bitmap []uint64, x []uint32, pos int, mask uint32) int
+
+//go:noescape
+func _hasAVX512() bool
+
+// AVX-512 is used only when the operating system saves the ZMM state and the
```
lemire, line 12: Yeah. This will be a problem, however.

lemire:
@perfloop-agent I found something better. Perfloop is credited 

https://github.com/RoaringBitmap/roaring/pull/557

perfloop-agent:
Addressed in `74b708afddae`. Response to [this feedback](https://github.com/RoaringBitmap/roaring/pull/556#issuecomment-5430400818).

The repaired source keeps the same block-wise compress-store approach and rebuilds it on the current base. Dense containers use the AVX-512 path, while sparse and unsupported targets retain the scalar fallback.

fillbits_avx512_amd64.go
perfloop-agent: Addressed in `74b708afddae`. Response to [this feedback](https://github.com/RoaringBitmap/roaring/pull/556#discussion_r3866370382).

The repaired source retains the OS and CPU feature gate before executing AVX-512 instructions, and keeps the ordered `ToArray` cursor and existing allocation path unchanged. This limits the new path to supported dense inputs.

perfloop-agent:
Thanks, Daniel. Agreed. The AVX512F gate did not answer your concern, and #557 is the better implementation. Sorry for the stale automated reply above. Closing #556 in favor of #557. Thank you for the Perfloop credit.

PR #573, opened 2026-09-19
roaring64: preserve the sign plane when cloning BSIs

jakezwang:
## Description

`NewBSIRetainSet` skips the sign plane because `BitCount()` excludes it. Since `Clone` delegates to this method, both operations turn negative BSI values into positive values. Include the sign plane in the existing copy-and-intersect loop.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Added public API regression coverage for signed int64 and wider values, zero/positive controls, retained membership and copy independence. The negative-value cases fail before this change.

Passed on macOS arm64:
- Go 1.24.13: `go test ./...` and `go test -tags appengine`
- Go 1.26.6: full `roaring64` and `BitSliceIndexing` tests with `-race`, focused copy tests with `-race`, `make unconvert`
- `gofmt` and `git diff --check`

The template's corpus generation command passes. Its native `-fuzz=FuzzSmat` command exits immediately: the current tree has no native fuzz target (`go test -list '^Fuzz'` lists none), so this does not count as a fuzz run. A separate temporary native fuzz test exercised valid public BSI copies against their input values, signs, membership and copy independence for 10 seconds (246,347 executions), with no failure.

The root package race run fails in unchanged `TestFrozenFormat` with a checkptr error at `serialization_littleendian.go:149`. The same `TestFrozenFormat` checkptr failure also reproduces on the pristine base (in its frozen-view conversion at line 137); the full race suite is not claimed as passing.

## Performance Impact

`BenchmarkBSIClone` measures a BSI with 1,000 mixed positive/negative values. Go 1.26.6 on macOS arm64, median of three 200ms runs:

```text
Before: BenchmarkBSIClone-8  14304 ns/op  16040 B/op  133 allocs/op
After:  BenchmarkBSIClone-8  16829 ns/op  17432 B/op  146 allocs/op
```

Copying the previously omitted sign plane adds its bitmap storage and one worker (+1,392 B/op, 13 allocations).


guymolinari:
Daniel,

I can also confirm that this is reproducible with v2.25.0.   I haven't merged but if you approve please merge ASAP.

Guy

